PR #23, opened 2020-11-02
Allow multiple JVM targets in kotlin-multiplatform projects

ALikhachev:
This change allows projects with multiple JVM targets to use the plugin. Required for adding JVM IR target in Kotlinx train build.

### DIFF
--- a/src/main/kotlin/ApiCompareCompareTask.kt
+++ b/src/main/kotlin/ApiCompareCompareTask.kt
@@ -11,6 +11,8 @@ import org.gradle.api.file.*
 import org.gradle.api.tasks.*
 import java.io.*
 
+const val COMPARE_TASk_BASE_NAME = "apiCheck"
+
 open class ApiCompareCompareTask : DefaultTask() {
     @InputDirectory
     @PathSensitive(PathSensitivity.RELATIVE)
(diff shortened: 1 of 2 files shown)


src/main/kotlin/ApiCompareCompareTask.kt
```diff
@@ -11,6 +11,8 @@ import org.gradle.api.file.*
 import org.gradle.api.tasks.*
 import java.io.*
 
+const val COMPARE_TASk_BASE_NAME = "apiCheck"
```
qwwdfsad, line 14: `COMPARE_TASk_BASE_NAME` -> `COMPARE_TASK_BASE_NAME`

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -119,11 +119,12 @@ private fun Project.configureApiTasks(sourceSet: SourceSet) {
...
 private fun Project.configureCheckTasks(
     apiBuildDir: File,
-    apiBuild: TaskProvider<KotlinApiBuildTask>
+    apiBuild: TaskProvider<KotlinApiBuildTask>,
+    targetName: String = ""
 ) {
     val projectName = project.name
-    val apiCheckDir = file(projectDir.resolve(API_DIR))
-    val apiCheck = task<ApiCompareCompareTask>("apiCheck") {
+    val apiCheckDir = file(projectDir.resolve(API_DIR + targetName.capitalize()))
```
qwwdfsad, line 126: It seems like the `/` is missing between `API_DIR` and `targetName`, isn't it?

qwwdfsad (changes requested):
The final question is: do we really have to maintain api dumps with the unstable compiler in stable libraries (where failed api dump fails all our builds and consumes our time)?
Can we avoid it?

ALikhachev:
During private discussion we've decided to try another way, these changes are no longer required.

PR #27, opened 2020-12-01
Support multiplatform when android is only JVM target

twyatt:
Adds support for `kotlin("multiplatform")` projects configured with `android()` as the only JVM target, for example:

```kotlin
plugins {
    id("com.android.library")
    kotlin("multiplatform")
}

kotlin {
    android()
    js().browser()

    // ...
}
```

_`jvm` platform type is favored when both `jvm()` and `android()` targets are present._


qwwdfsad (approved):
Thanks for your contribution! Looks good to me

I'll wait for #28 to be merged and to have a test infra to write a test for this change and then will merge it.

qwwdfsad:
Apparently, we do not have Android SDK on TeamCity, so merging it as is :) 
Thanks!

Issue #21, opened 2020-09-30
Build fails if api is not dumped


altavir:
The build fails with non-informative error (directory not found) if the initial API dump is not done before the build.
I think that the default behavior should be to give a warning and ignore the plugin altogether.

qwwdfsad:
Skipping the validation as the default behaviour for an applied plugin is a no-go. If skipping is the desired behaviour, the plugin shouldn't have been applied in the first place.

Added clear exception message instead

PR #28, opened 2020-12-01
Introduce 'ignoredClasses' configuration to exclude classes by fully …

### LINKED ISSUE #25: introduce ignoredClasses in gradle plugin
Type: Feat. request

Issue:
Android-Projects create a `BuildConfig` for android-libraries, which is public. It is not possible to exclude the package, because other needed classes are in given package. Also it is not possible to use an annotation, because it is generated.

Suggestion:
Introduce new `ignoredClasses` flag in gradle

Code
```
ignoredClasses += ["com.my.package.MyClazz"]
```

Note:
When this feat. is approved, I would implement it quickly and create a PR.

SergejIsbrecht:
Fixes #25

Open Points: 
* which code formatter should I use?
_should I move the functional-Tests to a different source-set?_

I touched a public API, which is used in Kotlin-STD-Lib. I think the change is breaking (byte-code), but when updating to another version, the calling could must only re-compile against the newest version. It should re-compile without issues, because a default value is provided.

```
@ExternalApi
public fun List<ClassBinarySignature>.filterOutNonPublic(nonPublicPackages: Collection<String> = emptyList(), nonPublicClasses: Collection<String> = emptyList()): List<ClassBinarySignature> {
``` 

Edit:
I moved the Gradle-Plugin functionalTests to their own SourceSet. Test will be run, when `gw test` is executed.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -21,6 +21,24 @@ sourceSets {
     }
 }
 
+sourceSets {
+    create("functionalTest") {
+        withConvention(org.jetbrains.kotlin.gradle.plugin.KotlinSourceSet::class) {
+        }
+        resources {
+            srcDir(file("src/functionalTest/resources"))
+        }
+        compileClasspath += sourceSets.main.get().output + configurations.testRuntimeClasspath
+        runtimeClasspath += output + compileClasspath
+    }
+}
+
+tasks.register<Test>("functionalTest") {
+    testClassesDirs = sourceSets["functionalTest"].output.classesDirs
+    classpath = sourceSets["functionalTest"].runtimeClasspath
+}
+tasks.check { dependsOn(tasks["functionalTest"]) }
+
 dependencies {
     implementation(gradleApi())
     implementation(kotlin("stdlib-jdk8"))
@@ -31,6 +49,10 @@ dependencies {
     implementation("com.googlecode.java-diff-utils:diffutils:1.3.0")
     compileOnly("org.jetbrains.kotlin.multiplatform:org.jetbrains.kotlin.multiplatform.gradle.plugin:1.3.61")
     testImplementation(kotlin("test-junit"))
+
+    "functionalTestImplementation"("org.assertj:assertj-core:3.18.1")
+    "functionalTestImplementation"(gradleTestKit())
+    "functionalTestImplementation"(kotlin("test-junit"))
 }
 
 tasks.withType<KotlinCompile>().configureEach {
@@ -77,6 +99,8 @@ extensions.getByType(PluginBundleExtension::class).apply {
 }
 
 gradlePlugin {
+    testSourceSets(sourceSets["functionalTest"])
+
     plugins {
         create("binary-compatibility-validator") {
             id = "org.jetbrains.kotlinx.binary-compatibility-validator"
--- a/src/main/kotlin/ApiValidationExtension.kt
+++ b/src/main/kotlin/ApiValidationExtension.kt
@@ -28,4 +28,10 @@ open class ApiValidationExtension {
      * Example of such annotation could be `kotlinx.coroutines.InternalCoroutinesApi`.
      */
     public var nonPublicMarkers: MutableSet<String> = HashSet()
+
+    /**
+     * Fully qualified names of classes that are ignored by the API check.
+     * Example of such a class could be `com.package.android.BuildConfig`.
+     */
+    public var ignoredClasses: MutableSet<String> = HashSet()
 }
--- a/README.md
+++ b/README.md
@@ -51,7 +51,8 @@ The plugin provides two tasks:
 
 Binary compatibility validator can be additionally configured with the following DSL:
 
-```kotlin
+Groovy
+```groovy
 apiValidation {
     /**
      * Packages that are excluded from public API dumps even if they
@@ -63,7 +64,13 @@ apiValidation {
      * Sub-projects that are excluded from API validation 
      */
     ignoredProjects += ["benchmarks", "examples"]
-    
+
+    /**
+     * Classes (fully qualified) that are excluded from public API dumps even if they
+     * contain public API.
+     */
+    ignoredClasses += ["com.company.BuildConfig"]
+
     /**
      * Set of annotations that exclude API from being public.
      * Typically, it is all kinds of `@InternalApi` annotations that mark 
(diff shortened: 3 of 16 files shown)


src/functionalTest/kotlin/kotlinx/validation/GradlePluginFuncTest.kt
```diff
@@ -0,0 +1,144 @@
...
+    @Rule
+    @JvmField
+    val testProjectDir: TemporaryFolder = TemporaryFolder()
+    private lateinit var buildFile: File
+    private lateinit var javaSourceFolder: File
+    private lateinit var apiDump: File
+
+    @Before
+    fun setup() {
+        apiDump = testProjectDir.newFolder("api")
```
qwwdfsad, line 28: It seems like a lot of ceremonies for the test. It would be much better to have a single test project already in the `resources` folder and use GradleRunner runner against it.

In addition to better readability, it will also significantly simplify writing tests in the future.
Your contribution is valuable as is, so feel free to leave it for me if you don't have enough time

build.gradle.kts
```diff
@@ -21,6 +21,24 @@ sourceSets {
     }
 }
 
+sourceSets {
```
qwwdfsad, line 24: Could you please elaborate on what's the purpose of a separate source set here?
E.g. why the tests cannot be located in a regular `test` package?

qwwdfsad (changes requested):
Great work, thanks for your contribution!

I have a few comments, let's address them and it's good to go

src/main/kotlin/ApiValidationExtension.kt
```diff
@@ -28,4 +28,10 @@ open class ApiValidationExtension {
      * Example of such annotation could be `kotlinx.coroutines.InternalCoroutinesApi`.
      */
     public var nonPublicMarkers: MutableSet<String> = HashSet()
+
+    /**
+     * Fully qualified names of classes that are ignored by the API check.
+     * Example of such a class could be `com.package.android.BuildConfig`.
+     */
+    public var ignoredClasses: MutableSet<String> = HashSet()
```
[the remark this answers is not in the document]
SergejIsbrecht: Yeah, will do. I somehow forgot.

build.gradle.kts
SergejIsbrecht: Well, the Gradle tasks are very slow, therefore the Gradle-Docs suggest, that you create a new SourceSet and run those tests on 'check'. You could get quick feedback via ./gradlew test, because the unit-tests run very fast. If you want to see, whether you broke something in the Gradle-Plugin bootstrapping, you could just run ./gradlew check in order to run the unit-tests and the Gradle-"integTests"

src/functionalTest/kotlin/kotlinx/validation/GradlePluginFuncTest.kt
SergejIsbrecht: I can do it, if you could explain how you would change it. The thing is, that testing the plugin is only possible with the Gradle-TestKit, which takes some setting up. Let's say you want to check different states of your Plugin. For example verify that ignored classes are excluded from apiDump. If you have no api folder, current implementation will crash, this is probably related to #21 . Therefore you have to make sure, that everything is in place, so the plugin is happy to execute given order. 

I would separate the unit-tests and the integration-Tests, that the plugin does, what it should be. You do not have to re-validate, that your business-logic works as expected in the integration-Tests (Gradle), but you have to make sure, that the plugin does what is has been told (e.g. dump the api, ignore files, ignore packages, does not fail if random error occurs. I would say the integration-tests for Gradle are kind of responsible for testing your "front-end" to your business-logic. Sure, it takes some setting up, but I do not see how this could be reduced.

src/main/kotlin/ApiValidationExtension.kt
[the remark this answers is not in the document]
SergejIsbrecht: Should I also add an example for kts? I already did, just asking whether this is needed or not, because you can not just use the groovy syntax to configure the plugin in kotlin.

build.gradle.kts
qwwdfsad: Ok, thanks for the explanation

src/main/kotlin/ApiValidationExtension.kt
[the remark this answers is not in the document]
qwwdfsad: Kts addition is very welcomed and definitely won't hurt

src/functionalTest/kotlin/kotlinx/validation/GradlePluginFuncTest.kt
qwwdfsad: Sure. Please take a look at how it is done in our atomicfu library as an example: https://github.com/Kotlin/kotlinx.atomicfu/commit/3b76f6dd52475b0d82ceb9c6018b990408abb99b

>Sure, it takes some setting up, but I do not see how this could be reduced.

For me, it looks like the following:

1) Let's write a test project as a regular folder structure and put it into the resources
2) The test copies the resources to a temporary directory and invokes `Gradle-TestKit` as it's currently being done
3) If you need to "tweak" the settings of an existing test project (e.g. to change `ApiValidationExtension`), instead of appending raw strings to test files, it can be done in the test project itself by using system properties or env. variable.


The idea is that writing/copy-pasting a small project is much more maintainable as opposed to juggling with raw strings: easy to edit, easy to read, hard to make a mistake (that is easy to when operating string literals). Also, such a structure is very welcoming for new tests: copy-paste the project, tweak its build file and you are already there!

SergejIsbrecht: I am currently looking into it. Sorry for the delay, I have been quite busy at work.

qwwdfsad: Sure, take your time, thanks for keeping me updated!
DSL looks good to me

SergejIsbrecht: Please have a look, I updated the PR according to given DSL. Some comments are missing and some proper error-handling, when the test-API is used not properly, but I think it can be used in order to write further tests.

qwwdfsad (approved):
Amazing work, thanks!

PR #29, opened 2020-12-01
Use defaultPublishConfig if possible for android modules

### LINKED ISSUE #24: Unable to apply plugin to flavored modules
Looks like the plugin only gets applied to Android modules if they have a [compilationTarget matching `release`](https://github.com/Kotlin/binary-compatibility-validator/blob/6b19ba4a3abfa70c5906750f747bff1e0d4faa76/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt#L52).

This makes it unusable if you have library flavors. Are there any plans to support multi flavored modules in the future? Would also be nice if you could notice this limitation in the docs.

pavelreiter:
Fixes #24 

I have prepared fix for #24 by adding compileOnly dependency on android gradle plugin (`com.android.tools.build:gradle:3.6.4`) and using android project extension to fetch [defaultPublishConfig](https://google.github.io/android-gradle-dsl/3.4/com.android.build.gradle.BaseExtension.html#com.android.build.gradle.BaseExtension:defaultPublishConfig) as preferred source of `compilationName` match.

### DIFF
--- a/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
+++ b/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
@@ -49,9 +49,9 @@ class BinaryCompatibilityValidatorPlugin : Plugin<Project> {
         project.pluginManager.withPlugin("kotlin-android") {
             if (project.name in project.rootProject.ignoredProjects()) return@withPlugin
             val extension = project.extensions.getByName("kotlin") as KotlinAndroidProjectExtension
-            val targetedPublications = project.targetedCompilation(extension.target.platformType)
+            val compilation = project.targetedCompilation(extension.target.platformType)
             extension.target.compilations.matching {
-                it.compilationName == targetedPublications
+                it.compilationName == compilation
             }.all {
                 project.configureKotlinCompilation(it, useOutput = true)
             }
@@ -64,9 +64,9 @@ class BinaryCompatibilityValidatorPlugin : Plugin<Project> {
                 it.platformType == KotlinPlatformType.jvm || it.platformType == KotlinPlatformType.androidJvm
             }.all { target ->
                 val useOutput = target.platformType == KotlinPlatformType.androidJvm
-                val targetedCompilation = project.targetedCompilation(target.platformType)
+                val compilation = project.targetedCompilation(target.platformType)
                 target.compilations.matching {
-                    it.name == targetedCompilation
+                    it.name == compilation
                 }.all {
                     project.configureKotlinCompilation(it, useOutput = useOutput)
                 }
--- a/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
+++ b/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
@@ -11,6 +11,7 @@ import org.gradle.api.tasks.*
 import org.jetbrains.kotlin.gradle.dsl.*
 import org.jetbrains.kotlin.gradle.plugin.*
 import java.io.*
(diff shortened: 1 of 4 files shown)


src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -48,8 +49,10 @@ class BinaryCompatibilityValidatorPlugin : Plugin<Project> {
         project.pluginManager.withPlugin("kotlin-android") {
             if (project.name in project.rootProject.ignoredProjects()) return@withPlugin
             val extension = project.extensions.getByName("kotlin") as KotlinAndroidProjectExtension
+            val androidBase = project.extensions.getByName("android") as AndroidBaseExtension
+            val defaultPublishName = androidBase.defaultPublishConfig ?: "release"
```
qwwdfsad, line 53: It seems like `defaultPublishConfig` is obsolete and recommended way to publish an application with `maven-publish` plugin is not using it: https://developer.android.com/studio/build/maven-publish-plugin.

I see the alternative solution where the user explicitly states the publication name they want to verify, then, if it's missing, we can lookup `defaultPublishConfig` and then fallback to release. It would be nice to have a test for that. 

The project doesn't have the infrastructure for tests yet (it is coming in #28), so the most valuable contribution would be to just put a bunch of sample Android projects into the resources: with legacy maven plugin, with `maven-publish` and with library flavours and I will take care of the rest.

qwwdfsad (changes requested):
Thanks

kingargyle:
Is there any update on this PR on when it might be addressed.   We are looking at using this for our android applications, but will need not only non-release support but flavor support as well.   Ideally, the apiDump should support something like, apiDumpDebug, apiDumpFlavorDebug, apiDumpRelease, apiDumpFlavorRelease.

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
pavelreiter: Hi, thanks for the comments.
I will try to look into it and since #28 is merged I will try to include some tests.

SumeraMartin:
@pavelreiter @qwwdfsad Hi, are you planning to finish this PR? Or do you have any workaround?

pavelreiter:
Hi, I am afraid that I have went with metalava for now. I will close the PR.

SumeraMartin:
I found a hacky workaround. Since I'm using flavors only for changes directly in the code `when (BuildConfig.FLAVOR) { ... }` and I don't use flavor-specific resources or classes then I'm disabling flavors in build.gradle.kts in case that apiCheck or apiDump task is being executed and I'm adding buildConfigField for FLAVOR with some dummy value. The code will compile fine, the task will succeed and flavors are still in the use. As I said, it is more hack than the correct solution but it works so far.

```
fun Project.startArgsContains(vararg arguments: String): Boolean {
    val startArgs = project.gradle.startParameter.taskRequests.toString()
    return arguments.any { argument -> startArgs.contains(argument) }
}

android {
    val isKotlinBinaryValidatorTask = project.startArgsContains("apiCheck", "apiDump")

    buildTypes {
        getByName(ProjectSettings.BuildType.DEBUG) {
            ...

            if (isKotlinBinaryValidatorTask) {
                buildConfigField("String", "FLAVOR", "\"dummy\"")
            }
        }
        ...
    }
    
    if (!isKotlinBinaryValidatorTask) {
        flavorDimensions(ProjectSettings.Flavor.DIMENSION)
        productFlavors {
            ...
        }
    }
    ...
}

PR #30, opened 2020-12-09
Properly dump $default counterparts of @PublishedApi methods

qwwdfsad:


ilya-g (approved):
Pretty hacky, IMO, but if tests pass then ok.

qwwdfsad:
Honestly, I have the same feeling.
Will be happy to rewrite it if a saner way is suggested

PR #31, opened 2020-12-14
Migrate to maven central publishing

qwwdfsad:
* Update Gradle to 6.7.1 because it supports in memory signing


buildSrc/src/Publishing.kt
```diff
@@ -0,0 +1,48 @@
...
+    }
+}
+
+fun PublishingExtension.mavenRepositoryPublishing() {
+    repositories {
+        maven {
+            url = URI("https://oss.sonatype.org/service/local/staging/deploy/maven2/")
+            credentials {
+                username = System.getenv("libs.sonatype.user")
+                password = System.getenv("libs.sonatype.password")
```
ilya-g, line 32: I believe it's better to use system properties instead of environment variables to pass values from teamcity to build. TeamCity system properties are available in Gradle as project properties. 
For the sensitive properties like these, I'd consider going even further: leaving them configuration only and passing to the single gradle step explicitly (though there's not much difference if a configuration consists of a single step).

qwwdfsad: Good points! I've discussed it with @goodwinnk and we concluded the following:
1) Env variable are generally safer because they are never preserved in the execution history and various JB builds are mostly using them
2) Narrowing the scope is indeed a good idea for deploy configuration that we are going to address systematically on TC configuration level.

To make it future-proof, I replaced `foo = System.getenv("foo")` with `foo = project.findProperty("foo") as? String ?: System.getenv("foo")` consistently

ilya-g (approved):

Issue #37, opened 2021-01-19
Allow applying the plugin to a subproject


sedovalx:
At the moment, the plugin must be applied to the root project. Otherwise, you get `Extension of type 'ApiValidationExtension' does not exist` error during the root project configuration. In my case, there is a Gradle multi-project with a hierarchy of hundreds of sub-projects and I want to validate only a single one. It seems to be quite natural to apply the plugin only to this single subproject.

As a workaround, I can apply the plugin to the root project and add everything to the `ignoredProjects` except the needed project. It works but doesn't look nice.

PR #39, opened 2021-01-27
Support applying the plugin to any project in hierarchy

joffrey-bion:
This change removes accesses to the extension via `rootProject`.
Instead, we now access the extension wherever it was actually created.

This means that when applying the plugin to a project that is not the rootProject, we will use the correct extension instead of looking for it in the rootProject (where it is not).

Resolves:
https://github.com/Kotlin/binary-compatibility-validator/issues/37


joffrey-bion:
This PR depends on the test facitlities added in PR #40, so please review/merge that one first, so that the diff of this PR is easier to read 😄

joffrey-bion:
@qwwdfsad thanks for merging #40 !

There was some merge issue during that merge because of a conflict with a [previous commit](99f75100e674378890e37039ad00ba42a50adaa9) that fixed a typo in the same test file that was renamed in #40.

This is why I deleted the extra file `IgnoredClassesTest.kt` in this PR. Note that the tests in this file are already present in `IgnoredClassesTests.kt` and `DefaultConfigTests.kt`, so no test was lost in the process.

qwwdfsad (approved):
👍

PR #45, opened 2021-02-20
Use Kotlin 1.4 in API dump tests 

ilya-g:
... and update public API dump after [KT-11567](https://youtrack.jetbrains.com/issue/KT-11567) codegen changes came into effect

The only questionable case left is a private companion in a public interface, which is generated as a public synthetic field.


qwwdfsad (approved):
👍

PR #50, opened 2021-03-18
Support Gradle's configuration cache

3flex:
I had a little trouble testing to see if I was successful, but this should allow builds using the Gradle plugin to take advantage of Gradle's [configuration cache](https://docs.gradle.org/current/userguide/configuration_cache.html) by removing direct access to `project` during task execution.


qwwdfsad (approved):
Thanks!

I'll merge it as is and then will also fix `projectName` -- it doesn't have to be `@Input` as it's constant through all the execution

Issue #20, opened 2020-08-25
Allow to dump public field nullability


e5l:
It will help us to track binary & source compatibility

qwwdfsad:
Changing nullability is always a source-incompatible change, always changes the runtime behaviour, but never breaks binary compatibility (no unsatisfied linkage).

Adding this feature is TBD

qwwdfsad:
For now, closing as won't fix

PR #51, opened 2021-03-28
Add nullability support

sebastianludwig:
Hey, we're currently in a situation where we need API compatibility above binary compatibility. BCV is already a pretty good fit for our needs, but nullability support would make it near perfect. So I took a stab at #20. The general approach is to add a `?` to every nullable type: `Ljava/lang/Integer` becomes `Ljava/lang/Integer?`.

Please let me know what you think about this approach. Also since Kotlin isn't my main language any stylistic feedback is also appreciated 😅

Open TODOs:
- [x] Since this is a breaking change we should make it opt-in and add an option to activate it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,12 @@ apiValidation {
      * Flag to programmatically disable compatibility validator
      */
     validationDisabled = true
+
+    /**
+     * Enable nullability annotations.
+     * Effectively appending `?` to nullable Types. Disabled by default.
+     */
+    annotateNullability = true
 }
 ```
 
@@ -119,6 +125,12 @@ apiValidation {
      * Flag to programmatically disable compatibility validator
      */
     validationDisabled = false
+
+    /**
+     * Enable nullability annotations.
+     * Effectively appending `?` to nullable Types. Disabled by default.
+     */
+    annotateNullability = true
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -201,6 +201,7 @@ For a class member a binary incompatible change is:
  - changing its name
  - changing its descriptor (erased return type and parameter types for methods);
    this includes changing field to method and vice versa
+ - changing the nullability/optionality of a type 
  - changing one of the following access flags:
     - `ACC_PUBLIC`, `ACC_PROTECTED`, `ACC_PRIVATE` — lessening the member visibility
     - `ACC_FINAL` — making non-final field or method final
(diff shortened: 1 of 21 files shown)


sebastianludwig:
I've added the option to enable nullability annotations in d61f33e. The diff would have been smaller if the parameter would be passed to `loadApiFromJvmClasses()` and from there to `toMethodBinarySignature()` but because the annotation is only applied when dumping the API I found it cleaner to reflect that in the API.

ilya-g:
Binary API dumps do not solve the problem of ensuring source compatibility, and missing nullability is not the only obstacle for it.

sebastianludwig:
Very much agreed. Do you think that adding nullability annotations is a valuable step towards verifying source compatibility or are you saying that attempting to derive a statement about source compatibility from API dumps is futile and thus this PR and #20 should be closed?

elizarov:
IMO, the goal of ensuring full source compatibility on changes is, in general, futile in a language like Kotlin. However, maintaining binary compatibility is extremely important for the libraries' ecosystem.

qwwdfsad:
In short, almost any source-level change that affects public signatures is source-incompatible.

For the two most basic scenarios (that covers 90+% of all changes) of signature changes it's easy to craft an example of a user's code that will break on a source level:

1) Adding a new public signature, e.g. `library.foo`. Users might have already had `usercode.foo` in their own codebase and two wildcard imports -- one for `library.*` and one for `usercode.*`. After our addition of **new** API, all calls to `foo` in such places will be ambiguous, thus adding new API is source-incompatible.

2) Changing the public signature is basically the same, but instead of being ambiguous, our modified signature might become more or less specific than user's existing ones (or equal to it!), thus also changing a behaviour.

So, either the tool will complain on _any_ change (then it has no uses at all), or it will be unsound and let some incompatible changes slip in as compatible, which also is a no go

sebastianludwig:
> So, either the tool will complain about any change (then it has no uses at all)

Well, that's was our use case 😅 We were looking for a tool to diff our API to the previous release to ensure that we didn't overlook any sneaky `?` in our code reviews and mentioned everything in the changelog.

joffrey-bion:
> In short, almost any source-level change that affects public signatures is source-incompatible

@qwwdfsad True, but I don't see how this means that we wouldn't we want an automated tool to check for changes in public signatures. It could help decisions about version increments etc. AND it could be multiplatform.

This is one of my usages of `binary-compatibility-validator`, but it's not the correct tool for the job and it's JVM-only (it doesn't care about changes in Native-specific code for instance), which makes sense given its purpose.

PR #59, opened 2021-06-02
Evaluate 'static final Companion' visibility in the outer class to th…

### LINKED ISSUE #53: The Companion field corresponding to an internal companion object is not part of the public API
If I have a class declared like:

```kt
package pbandk

class Foo internal constructor() {
    internal companion object {
        ...
    }
}
```

The API dump is incorrectly showing the `Companion` field as part of the public API:

```
public final class pbandk/Foo {
	public static final field Companion Lpbandk/Foo$Companion;
}
```

qwwdfsad:
…e visibility of the companion class itself

Fixes #53

### DIFF
--- a/src/main/kotlin/api/KotlinSignaturesLoading.kt
+++ b/src/main/kotlin/api/KotlinSignaturesLoading.kt
@@ -10,6 +10,7 @@ import kotlinx.validation.*
 import org.objectweb.asm.*
 import org.objectweb.asm.tree.*
 import java.io.*
+import java.util.*
 import java.util.jar.*
 
 @ExternalApi
@@ -27,9 +28,11 @@ public fun Sequence<InputStream>.loadApiFromJvmClasses(visibilityFilter: (String
         }
     }
 
-    val visibilityMap = classNodes.readKotlinVisibilities().filterKeys(visibilityFilter)
-    val classNodeMap = classNodes.associateBy { it.name }
-    return classNodes
+    // Note: map is sorted, so the dump will produce stable result
+    val classNodeMap = classNodes.associateByTo(TreeMap()) { it.name }
+    val visibilityMap = classNodeMap.readKotlinVisibilities(visibilityFilter)
+    return classNodeMap
+        .values
         .map { classNode ->
             with(classNode) {
                 val metadata = kotlinMetadata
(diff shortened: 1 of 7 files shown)


src/test/kotlin/cases/internal/internal.txt
```diff
@@ -1,3 +1,16 @@
+public final class cases/internal/Companion {
+	public fun <init> ()V
+}
+
+public final class cases/internal/NameShadowing {
+	public fun <init> ()V
+	public static final fun getCompanion ()Lcases/internal/Companion;
```
qwwdfsad, line 7: This test exposes another problem: `ACC_PUBLIC` method lifted to the outer class via `@JvmStatic` are considered to be a public API.

This case seems to be pretty rare, so if you don't mind, I'd like to postpone the fix of this problem

src/main/kotlin/api/KotlinSignaturesLoading.kt
```diff
@@ -26,19 +27,33 @@ public fun Sequence<InputStream>.loadApiFromJvmClasses(visibilityFilter: (String
         }
     }
 
-    val visibilityMapNew = classNodes.readKotlinVisibilities().filterKeys(visibilityFilter)
+    val visibilityMap = classNodes.readKotlinVisibilities().filterKeys(visibilityFilter)
+    val classNodeMap = classNodes.associateBy { it.name }
```
ilya-g, line 31: I think it's time to do something with repeated re-reading of classNodes - it's a sequence that reads all classes again each time.
Perhaps associate them by name first, then use `classNodeMap.values` instead of `classNodes`. If that map is sorted, we can avoid sorting these values by name in the end.

src/test/kotlin/cases/internal/internalCompanion.kt
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright 2016-2021 JetBrains s.r.o.
+ * Use of this source code is governed by the Apache 2.0 License that can be found in the LICENSE.txt file.
+ */
+
+package cases.internal
+
+// Internal companion is not part of public API
+// neither should be outer static final companion field
+class Companion {
```
ilya-g, line 10: Since they share package with the other test declarations, better to name them more specifically, e.g. `PublicClassInternalCompanion`. Or does it named so to ensure that the object `Companion` name is not confused with the class `Companion` name?

ilya-g (approved):

src/test/kotlin/cases/internal/internalCompanion.kt
qwwdfsad: >does it named so to ensure that the object Companion name is not confused with the class Companion name?

Right, dropped a comment, renamed non-relevant companions

PR #62, opened 2021-07-02
Fixed multiple jvm targets

pdvrieze:
This is a rework of the multiple-jvm-targets branch that actually works. The testing is "fixed" and the implementation is fixed. Note that this is not 100% compatible with existing projects as it uses target-specific subdirectories (only for multiplatform projects).

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -50,6 +50,8 @@ val createClasspathManifest = tasks.register("createClasspathManifest") {
     }
 }
 
+val kotlinVersion: String by project
+
 dependencies {
     implementation(gradleApi())
     implementation("org.jetbrains.kotlinx:kotlinx-metadata-jvm:0.3.0")
@@ -57,7 +59,7 @@ dependencies {
     implementation("org.ow2.asm:asm-tree:9.0")
     implementation("com.googlecode.java-diff-utils:diffutils:1.3.0")
     compileOnly("org.jetbrains.kotlin.multiplatform:org.jetbrains.kotlin.multiplatform.gradle.plugin:1.3.61")
-    testRuntimeOnly("org.jetbrains.kotlin.multiplatform:org.jetbrains.kotlin.multiplatform.gradle.plugin:1.5.20")
+    testRuntimeOnly("org.jetbrains.kotlin.multiplatform:org.jetbrains.kotlin.multiplatform.gradle.plugin:$kotlinVersion")
     testImplementation(kotlin("test-junit"))
     "functionalTestImplementation"(files(createClasspathManifest))
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -36,17 +36,29 @@ tasks.register<Test>("functionalTest") {
 }
 tasks.check { dependsOn(tasks["functionalTest"]) }
 
-// Hack (from guidance on gradle forums) needed to handle optional plugin dependencies
-// in test code. In test mode classloading works different and classes provided by other
-// plugins are not visible.
+// While gradle testkit supports injection of the plugin classpath it doesn't allow using dependency notation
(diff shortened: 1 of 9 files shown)


qwwdfsad:
Thanks! Currently I'm busy with other things, but I'll try to return to this PR by the end of this week

pdvrieze:
The pull request has been rebased upon 0.7.1

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -58,31 +59,58 @@ class BinaryCompatibilityValidatorPlugin : Plugin<Project> {
...
+fun KotlinTarget?.apiTaskName(suffix: String) = when (this?.name) {
+    null,
+    "jvm" -> "api$suffix"
+    else  -> "${name}Api$suffix"
+}
+
+val KotlinTarget?.apiDir
+    get() = when (this?.name) {
+        null -> API_DIR
+        else -> "$API_DIR/$name"
```
qwwdfsad, line 100: It would be nice to keep the BCV compatible and store `jvm` dump in a top-level `api` dir _unless_ there are multiple JVM targets

qwwdfsad (changes requested):
Sorry for the delay!

The build does not pass as the plugin now creates the same task twice for any multiplatform project with a single JVM source set, this definitely should not be the case. In fact, it's impossible to apply it to any multiplatform project

pdvrieze:
> Sorry for the delay!
> 
> The build does not pass as the plugin now creates the same task twice for any multiplatform project with a single JVM source set, this definitely should not be the case. In fact, it's impossible to apply it to any multiplatform project

This should be fixed in the new pull request.

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -58,47 +68,117 @@ class BinaryCompatibilityValidatorPlugin : Plugin<Project> {
...
+            }
+        } ?: API_DIR_PROVIDER
+
+}
+
+
+enum class DirConfig {
+    COMMON,
+    TARGET_DIR,
+    AUTO
```
qwwdfsad, line 145: AUTO seems to be unused in the code

qwwdfsad (commented):
Great job, thanks!

Please address few minor comments and then the PR will be ready to go: I will do a few minor stylistic cleanups and will release 0.8.0-RC next week

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
pdvrieze: The idea was to make this user configurable (in the configuration extension), but with the working count provider it is actually not needed anymore. I'll remove it.

build.gradle.kts
```diff
@@ -36,6 +36,9 @@ tasks.register<Test>("functionalTest") {
 }
 tasks.check { dependsOn(tasks["functionalTest"]) }
 
+// Hack (from guidance on gradle forums) needed to handle optional plugin dependencies
```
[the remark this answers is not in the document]
pdvrieze: Looking it up, there should actually be a better way: https://docs.gradle.org/current/userguide/test_kit.html#sub:test-kit-automatic-classpath-injection (rather than https://docs.gradle.org/2.14/userguide/test_kit.html?_ga=2.33896865.1010618029.1633210343-1565008009.1633210343#sub:test-kit-classpath-injection). I'll try to use the better way.

pdvrieze:
All three changes have been processed in the latest push. I found a better source for the code, and documented what is happening much better.

qwwdfsad (approved):
That's a great job, thanks for taking care of it!

PR #67, opened 2021-08-17
Fail-fast if it's not possible to read @Metadata from Kotlin class in…

qwwdfsad:
…stead of producing incorrect visibilities and dumps


src/main/kotlin/api/KotlinMetadataVisibilities.kt
```diff
@@ -60,6 +61,14 @@ val ClassNode.kotlinMetadata: KotlinClassMetadata?
             )
         }
         return KotlinClassMetadata.read(header)
+            ?: error(
+                """
+                Incompatible version of Kotlin metadata.
+                Maximal supported Kotlin metadata version: ${COMPATIBLE_METADATA_VERSION.joinToString(".")},
+                $name Kotlin metadata version: ${header.metadataVersion.joinToString(".")}.
+                As a workaround, it is possible to manually update 'kotlinx-metadata' version in your project.
```
udalov, line 69:
```suggestion
                As a workaround, it is possible to manually update 'kotlinx-metadata-jvm' version in your project.
```

udalov (approved):

PR #72, opened 2021-11-24
Synthetic annotations support

qwwdfsad:


build.gradle.kts
```diff
@@ -86,7 +86,7 @@ dependencies {
 
 tasks.withType<KotlinCompile>().configureEach {
     kotlinOptions.apply {
-        languageVersion = "1.4"
+        languageVersion = "1.6"
```
ilya-g, line 89: Beware that the Kotlin compiler below 1.5 won't be able to compile against classes with such metadata version. This may be a problem for a compiler bundled with Gradle, so the version of the compiler bundled into the minimal supported Gradle version should be considered.

src/main/kotlin/api/AsmMetadataLoading.kt
```diff
@@ -27,18 +27,19 @@ fun isStatic(access: Int) = access and Opcodes.ACC_STATIC != 0
...
             && !isWhenMappings()
+            && !isSyntheticAnnotationClass()
             && (classVisibility?.isPublic(isPublishedApi()) ?: true)
 
 
 val ClassNode.innerClassNode: InnerClassNode? get() = innerClasses.singleOrNull { it.name == name }
 fun ClassNode.isLocal() = outerMethod != null
 fun ClassNode.isInner() = innerClassNode != null
 fun ClassNode.isWhenMappings() = isSynthetic(access) && name.endsWith("\$WhenMappings")
+fun ClassNode.isSyntheticAnnotationClass() = isSynthetic(access) && name.contains("\$annotationImpl\$")
```
ilya-g, line 42: I don't mind detecting annotation implementation classes this way, but what bothers me is that these classes are public at all. At first sight, being package private would be enough for them.

src/main/kotlin/api/AsmMetadataLoading.kt
```diff
@@ -27,18 +27,19 @@ fun isStatic(access: Int) = access and Opcodes.ACC_STATIC != 0
 fun isFinal(access: Int) = access and Opcodes.ACC_FINAL != 0
 fun isSynthetic(access: Int) = access and Opcodes.ACC_SYNTHETIC != 0
 
-
 fun ClassNode.isEffectivelyPublic(classVisibility: ClassVisibility?) =
     isPublic(access)
             && !isLocal()
             && !isWhenMappings()
+            && !isSyntheticAnnotationClass()
```
ilya-g, line 34: Turns our that annotation classes can contribute to the public binary API due to potential unrestricted cross-module inlining, at least their constructors.
сс @sandwwraith

ilya-g (changes requested):

src/main/kotlin/api/AsmMetadataLoading.kt
qwwdfsad: They can be declared in `inline` functions and as long as they are synthetic, the compiler doesn't bother with detecting that and just makes it `public` (similarly to SAM-constructors)

src/main/kotlin/api/AsmMetadataLoading.kt
qwwdfsad: When inlining crossing the module boundaries, the compiler re-creates the class to avoid messing with ABI

build.gradle.kts
qwwdfsad: Good point, I've reverted that and applied 1.6.0 LM only to tests

Issue #76, opened 2022-04-01
Top-level projects don't seem to work on all OS's 


ZacSweers:
I'm not entirely sure if I'm reading this all right, but in our single-project repo doesn't appear to work when running `apiCheck` on linux CI while it does work on macOS locally.

Here's a repro - https://github.com/slackhq/EitherNet/pull/52. Fails on CI with a strange check that suggests it's expecting a subproject, but it's a single-project repo only.

joffrey-bion:
Why do you say it seems to be expecting a subproject? From your repro I could see:

```
Execution failed for task ':apiCheck'.
> File eithernet.api is missing from api/, please run :eithernet:apiDump task to generate one
```

Since you do have [api/EitherNet.api](https://github.com/slackhq/EitherNet/blob/main/api/EitherNet.api) (with CamelCase), I wonder if it's a case issue instead. Did `apiDump` generate the file with this camel case?

Maybe the bug would be due to an inconsistency between `.api` files generation and inspection. The former respecting the case of the project name, and the latter expecting all lowercase?

ZacSweers:
This only reproduces on Linux, nothing errors when trying this flow on macOS locally.

Look again at the message

```
please run :eithernet:apiDump task
```

This doesn't make sense because there _is_ no `eithernet` subproject. There is only one project — the root project. The only dump task is `:apiDump`, not `:eithernet:apiDump`. I think the problem lies somewhere in there - why does the plugin think that subproject exists?

ZacSweers:
It could be that it expects an all lowercase name on Linux, but I would expect the error message to be different or to possibly just better handle this case

joffrey-bion:
> This doesn't make sense because there is no eithernet subproject [...] I think the problem lies somewhere in there - why does the plugin think that subproject exists?

Oh right, the error message indeed looks broken in this respect. But looking at the code it really seems it is just that - a misleading error message. However, I'm not the maintainer of this project, so please take that with a grain of salt.

> This only reproduces on Linux, nothing errors when trying this flow on macOS locally.

Filename case sensitivity behaving differently on different systems wouldn't surprise me. Better have the correct case instead of relying on system quirks.

IMO the problem is that you ran `apiDump` with the default project name, and later added an explicit `rootProject.name` in the Gradle settings without regenerating the api files again.

[...]

PR #77, opened 2022-04-04
Fix task name in error message for missing .api files (and add some hints)

joffrey-bion:
Resolves the task name bug mentioned in this issue: https://github.com/Kotlin/binary-compatibility-validator/issues/76

And also provides some hints about project renames (when a file is found with an incorrect name) and suggest deletion in case the rename is case-change-only.

I'm sorry if the diff is a bit chaotic, I tried to simplify some stuff in the implementation that accesses the files. Not sure why the objectfactory / filetrees were used there instead of plain listFiles().

### DIFF
--- a/src/functionalTest/kotlin/kotlinx/validation/api/Assert.kt
+++ b/src/functionalTest/kotlin/kotlinx/validation/api/Assert.kt
@@ -5,6 +5,7 @@
 
 package kotlinx.validation.api
 
+import org.assertj.core.api.Assertions
 import org.gradle.testkit.runner.BuildResult
 import org.gradle.testkit.runner.TaskOutcome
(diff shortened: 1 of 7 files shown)


qwwdfsad (commented):
Build seems to fail, please fix corresponding tests

joffrey-bion:
> Build seems to fail, please fix corresponding tests

Sorry about that, it seems I updated one work in the error message after running the tests. Will fix ASAP

qwwdfsad:
Sorry, we've fixed it all separately and the changeset is no longer relevant.
All corresponding bugs are fixed though!

joffrey-bion:
Sure, no worries. If the task name thing was fixed too, it's alright then :)

PR #79, opened 2022-04-11
Look for .api file in a case-insensitive manner

### LINKED ISSUE #76: Top-level projects don't seem to work on all OS's

qwwdfsad:
Fixes #76

### DIFF
--- a/src/functionalTest/kotlin/kotlinx/validation/test/DefaultConfigTests.kt
+++ b/src/functionalTest/kotlin/kotlinx/validation/test/DefaultConfigTests.kt
@@ -90,6 +90,29 @@ internal class DefaultConfigTests : BaseKotlinGradleTest() {
     }
 
     @Test
+    fun `apiCheck should succeed when public classes match api file ignoring case`() {
+        val runner = test {
+            buildGradleKts {
+                resolve("examples/gradle/base/withPlugin.gradle.kts")
+            }
+            kotlin("AnotherBuildConfig.kt") {
+                resolve("examples/classes/AnotherBuildConfig.kt")
+            }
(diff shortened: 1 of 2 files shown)


ilya-g:
Maybe it's worth to log a warning when the case of an api file name doesn't match the expected case?
Could `apiDump` task be changed to normalize the case of the file name?

ilya-g (approved):

qwwdfsad:
Sorry, thought we've discussed it on meeting, should've written the rationale anyway.

The problem with mixed case sensitive/case-insensitive is that there is no single source of truth -- if the project is being developed exclusively on OS X, than CamelProjectCase can be preferable over normalized file name. Normalizing it will also trigger disruptive change for everyone updated, which isn't a good thing.

Also, the original issue (https://github.com/slackhq/EitherNet/pull/52) is really weird and it's hard to tell whether a single change in minor Gradle version leading to changing the case of project name in the Gradle model a desired behaviour or a bug and triggering a warning over such thing doesn't seem reasonable to me.

ZacSweers:
unfortunately it looks like it still fails in 0.9.0 :/. Same PR

ZacSweers:
part of the problem is that it uses the project name (which has capital letters). If I regenerate it with all lowercase, then it works. It does seem to expect it to be all lowercase though when checking

PR #80, opened 2022-05-16
Do not expose kotlin-stdlib in dependencies

martinbonnin:
Like for the Kotlin Gradle Plugin [YouTrack comment](https://youtrack.jetbrains.com/issue/KT-41142), there is no need to expose `kotlin-stdlib` as a runtime dependency as Gradle will provide its own version on the classpath. 

This fixes some `Runtime JAR files in the classpath should have the same version` warnings.


qwwdfsad (approved):
Thanks for taking care of it 👍

PR #82, opened 2022-05-16
Also exclude transitive usages of the stdlib

martinbonnin:
Follow up from https://github.com/Kotlin/binary-compatibility-validator/pull/80

It turns out the `stdlib` was also transitively pulled by `kotlinx-metadata-jvm` so I excluded it from the `implementation` configuration altogether.


qwwdfsad (approved):
Thanks!

PR #84, opened 2022-05-17
Fix case-sensitivity of FS lookup

### LINKED ISSUE #76: Top-level projects don't seem to work on all OS's

qwwdfsad:
Fixes #76


qwwdfsad:
All our hardware is moving, so I cannot validate the fix locally. But prior to the change, the test [has been failing](https://teamcity.jetbrains.com/viewLog.html?buildId=3855717&tab=buildResultsDiv&buildTypeId=KotlinTools_BinaryCompatibilityValidator_Build) on Linux agents

Issue #58, opened 2021-06-02
Non-public markers aren't handled correctly for functions with default parameters


zsmb13:
When declaring an opt-in annotation and adding that annotation to the `nonPublicMarkers` configuration, functions that have default parameters are not correctly excluded from the api dump - the generated `$default` function is still included. A very similar issue to the one fixed in https://github.com/Kotlin/binary-compatibility-validator/pull/30.

Example annotation:

```kotlin
@RequiresOptIn(level = RequiresOptIn.Level.WARNING)
annotation class MyInternalApi
```

An example function:

```kotiln
@MyInternalApi
fun defParamInternalFun(x: Int = 0) {}
```

The incorrect API dump:

```
public final class co/zsmb/example/MainKt {
	public static synthetic fun defParamInternalFun$default (IILjava/lang/Object;)V
}
```

Project with examples and reproducing the issue can be found [here](https://github.com/zsmb13/binary-compat-internal-annotation-default-param-issue/).

qwwdfsad:
Fixed in 0.10.0

PR #81, opened 2022-05-16
Track non-public annotations on properties

martinbonnin:
Attempt at fixing https://github.com/Kotlin/binary-compatibility-validator/issues/36

PR #85, opened 2022-05-19
For synthetic $default methods, lookup the annotations on their corresponding non-synthetic method

martinbonnin:
Follow up from https://github.com/Kotlin/binary-compatibility-validator/pull/81 as the logic is pretty similar

closes https://github.com/Kotlin/binary-compatibility-validator/issues/58 

PS: not sure why I disabled maintainer edits on that other PR but this one has the checkbox enabled ✅  so feel free to push here if needed


src/main/kotlin/api/KotlinSignaturesLoading.kt
```diff
@@ -85,7 +85,19 @@ public fun Sequence<InputStream>.loadApiFromJvmClasses(visibilityFilter: (String
                             ?.visibleAnnotations ?: emptyList()
                     }
 
-                    it.toMethodBinarySignature(foundAnnotations)
+                    /**
+                     * For synthetic $default methods, pull the annotations from the corresponding method
+                     */
+                    val alternateDefaultSignature = mVisibility?.name?.let { className ->
```
qwwdfsad, line 91: Would it be okay for you to take an extra step and consolidate your approach with the hackish filtering of `publishedApiSignatures` few lines below?

It seems like the handling is mostly identical and we have two similar codepaths, e.g. `isPublishedApiWithDefaultArguments` and `alternateDefaultSignature`

martinbonnin: I'll take a look!

src/test/kotlin/cases/markerDefault/marker.kt
```diff
@@ -0,0 +1,10 @@
+package cases.markerDefault
```
ilya-g, line 1: Test case directories are usually more coarse-grained, they contain tests for multiple related cases. So it would make sense to place this new case in the `marker` directory.

src/main/kotlin/api/KotlinSignaturesLoading.kt
martinbonnin: Refactor is pushed. It took a bunch of changes. I'm pretty sure this is sub-optimal in terms of performance (the `$annotations` method could potentially be looked up twice for field + getter). Let me know how much of an issue it is.

As a nice side effect, I could enable [all of the "inline" tests](https://github.com/Kotlin/binary-compatibility-validator/blob/b98502b1fc42eb7f98020ace39ede4eb8801acec/src/test/kotlin/cases/inline/inlineExposed.kt#L14)

martinbonnin: Something else I wanted to mention: the new changes now collect both visible + invisible annotations for markers (because `@PublishedApi` is invisible it seems). I'm not super versed in the visible vs invisible difference but this too might have a performance hit.

qwwdfsad: Performance doesn't seem to be an issue here; thanks for expressing your concerns though!

qwwdfsad (approved):
Superb, thanks! 👍

PR #96, opened 2022-07-25
Restore compatibility with Gradle's configuration cache

### LINKED ISSUE #95: apiDump task incompatible with configuration cache
```
FAILURE: Build failed with an exception.

* What went wrong:
Configuration cache problems found in this build.

10 problems were found storing the configuration cache.
- Task `:detekt-api:apiDump` of type `org.gradle.api.tasks.Sync`: cannot serialize object of type 'kotlinx.validation.KotlinApiBuildTask', a subtype of 'org.gradle.api.Task', as these are not supported with the configuration cache.
  See https://docs.gradle.org/7.5/userguide/configuration_cache.html#config_cache:requirements:task_access
- Task `:detekt-api:apiDump` of type `org.gradle.api.tasks.Sync`: cannot serialize object of type 'org.gradle.api.internal.project.DefaultProject', a subtype of 'org.gradle.api.Project', as these are not supported with the configuration cache.
  See https://docs.gradle.org/7.5/userguide/configuration_cache.html#config_cache:requirements:disallowed_types
```

3flex:
Fixes #95

### DIFF
--- a/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
+++ b/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
@@ -182,7 +182,7 @@ private fun Project.configureKotlinCompilation(
     val apiBuild = task<KotlinApiBuildTask>(targetConfig.apiTaskName("Build"), extension) {
         // Do not enable task for empty umbrella modules
         isEnabled =
-            apiCheckEnabled(extension) && compilation.allKotlinSourceSets.any { it.kotlin.srcDirs.any { it.exists() } }
+            apiCheckEnabled(projectName, extension) && compilation.allKotlinSourceSets.any { it.kotlin.srcDirs.any { it.exists() } }
         // 'group' is not specified deliberately, so it will be hidden from ./gradlew tasks
         description =
             "Builds Kotlin API for 'main' compilations of $projectName. Complementary task and shouldn't be called manually"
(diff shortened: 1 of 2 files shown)


src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -261,15 +261,12 @@ private fun Project.configureCheckTasks(
...
 
     val apiDump = task<Sync>(targetConfig.apiTaskName("Dump")) {
-        isEnabled = apiCheckEnabled(extension) && apiBuild.map { it.enabled }.getOrElse(true)
+        isEnabled = apiCheckEnabled(projectName, extension) && apiBuild.map { it.enabled }.getOrElse(true)
         group = "other"
         description = "Syncs API from build dir to ${targetConfig.apiDir} dir for $projectName"
         from(apiBuildDir)
         into(apiCheckDir)
         dependsOn(apiBuild)
-        doFirst {
```
qwwdfsad, line 270: Could you please elaborate on why you removed this `mkdirs` call?

Does `into` documentation guarantees `mkdir`?

qwwdfsad (commented):
Thanks for taking care of it!

I have a question about `mkdirs` and this PR is good to go

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
3flex: It's odd, as this doesn't seem to violate the configuration cache contract, but it wouldn't work unless I removed it.

In any case, yes, `into` will create the required directories. More here: https://docs.gradle.org/current/userguide/working_with_files.html#sec:creating_directories_example

> All core Gradle tasks ensure that any output directories they need are created if necessary

qwwdfsad: Works for me, thanks!
At least now we have a proper test coverage for that

qwwdfsad (approved):
Once again, thanks for taking care of it!

3flex:
Any chance we could get a new release cut? Not sure if there's anything else planned for the next version, but I'd really like to get rid of the `notCompatibleWithConfigurationCache` call for these tasks in the build scripts!

qwwdfsad:
0.11.1 is published to maven central. Should be available for regular dependencies within an hour

PR #98, opened 2022-09-18
Add Java Android Library Support

### LINKED ISSUE #94: Java Android Library Support
Binary compatibility plugin configures necessary tasks like 'apiDump' and 'apiCheck' only if `kotlin`, `kotlin-android` or `kotlin-multiplatform` plugins are applied as can be seen [here](https://github.com/Kotlin/binary-compatibility-validator/blob/89ad35ec75dcbb6ede9b72f7185416783828db02/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt#L112) 

Although it can be used for Java libraries, due to missing the plugins listed above, necessary tasks are not created.

Java Android libraries can be supported if the tasks are also created for projects where `com.android.library` plugin is applied.

realbarisbasturk:
Fixes #94

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -13,6 +13,7 @@ plugins {
 repositories {
     mavenCentral()
     gradlePluginPortal()
+    google()
 }
 
 sourceSets {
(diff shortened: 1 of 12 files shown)


qwwdfsad:
Great job! 

I cannot merge it as is as it requires ANDROID_SDK configuration on our TeamCity, which may take some time

qwwdfsad:
Thanks! Merged manually in https://github.com/Kotlin/binary-compatibility-validator/pull/102

PR #105, opened 2022-10-26
Revert 'com.android.library' support

qwwdfsad:


qwwdfsad:
See https://github.com/Kotlin/binary-compatibility-validator/issues/94#issuecomment-1291757630 for rationale

Issue #70, opened 2021-10-21
Is it possible to distinguish between compatible and incompatible changes? 


martinbonnin:
When monitoring public API, I can think of two use cases:

1. Make sure that `module.api` files are always up-to-date and represent the latest state of the public API.
2. Make sure that a new version of `module.api` is backward compatible with the previous one to avoid breaking changes.

In the case of 1., it's important to detect additions as we want them to be taken into account in subsequents checks. 

But additions are typically not important for 2.: It is fine to release a new version of a lib with new APIs as it will not break backward compatibility.

My current understanding is that `./gradlew apiCheck` is focused at 1. and will fail on any change, even additions. Am I correct? 

If that's the case, it would be nice to have another task focused at 2. that will only succeed if the new API is backward compatible. 

Maybe `./gradlew apiDumpSafe` ?

It would fail on backward incompatible changes only and output a report with the problematic changes and whether they are source-level or binary-level incompatibilities. 

Is that currently possible? Or will it be?

qwwdfsad:
Sorry for the extremely late reply.

Alas, I only see that such a mode may make the API changes less robust and easier to break.

When the authors use `apiDumpSafe` and `check` checks only safe API, nothing prevents API authors to release new API without updating `module.api` file to reflect the fact that any changes in these signatures are now considered breaking.

Additionally, such mode is not intended as the main usage pattern of this tool and having it operating in two modes means twice as much testing and maintenance. 

I can suggest an alternative -- it is possible to depend on `binary-compatibility-validator` as a JAR with public API (that mostly dumps `.api` files) and post-process the results in the way you need it

martinbonnin:
> When the authors use apiDumpSafe and check checks only safe API, nothing prevents API authors to release new API without updating module.api file to reflect the fact that any changes in these signatures are now considered breaking.

Not sure I follow... There would be 2 use cases:

1. library is stable. Maintainers call `apiDumpSafe`. Only additions are allowed.
2. library is not stable. Maintainers call `apiDump` to overwrite whatever API was there before. Breaking changes are allowed.

In all cases, the CI calls `apiCheck` and returns an error if the committed files does not match the API. 

Japicmp for an example has something like this where it shows what is breaking and what is non-breaking.

> and post-process the results in the way you need it

Sure. I'd argue my needs aren't that specific. So if that post processing could be in a central first-party place, that'd benefit the ecosystem as a whole but I can try to work on something at some point :)

[...]

PR #114, opened 2022-12-27
Add ability to turn on cheking only changes and deletes in api

jumpstreet13:
In our library, we want ability to turn on checking only delete and update changes (not creating new classes for example).
Added this ability to plugin


qwwdfsad:
This functionality was discussed multiple times: https://github.com/Kotlin/binary-compatibility-validator/issues/70 (and linked)

We are not ready to accept such a contribution. The API surface of BCV is extensive enough to emulate it on user side if there is need to

jumpstreet13:
Thought this would be a handy option for many who use your library
thanks anyway

PR #119, opened 2023-02-15
Update kotlinx-metadata-jvm to 0.6.0 and Kotlin to 1.8.10

sandwwraith:
Remove -Xskip-runtime-version-check flag as it is deprecated and has no effect.

Add -Xsuppress-version-warnings flag to suppress "LV 1.4 is deprecated" warning and enable allWarningsAsErrors back.

Remove unused declarations that contained deprecated code (.lowercase()).

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -74,11 +74,11 @@ configurations.implementation {
 
 dependencies {
     implementation(gradleApi())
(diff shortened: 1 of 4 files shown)


ilya-g (commented):
👍

sandwwraith:
@qwwdfsad Can you make a new release? An updated metadata version is required for switching Kotlin master to LV=1.9, and BCV should be updated too to work with the new metadata version.

PR #122, opened 2023-03-01
Update to Gradle 8

aSemy:
- bump Gradle Plugin Publish Plugin version
- build script updates & improvements
- use Java Toolchains to set Java version
- use jvm-test-fixtures plugin for functionalTest sources


build.gradle.kts
```diff
@@ -109,23 +86,23 @@ tasks.compileKotlin {
...
+tasks.compileTestKotlin {
+    kotlinOptions {
+        languageVersion = "1.6"
     }
 }
 
+tasks.withType<Test>().configureEach {
+    systemProperty("overwrite.output", System.getProperty("overwrite.output", "false"))
+    systemProperty("testCasesClassesDirs", sourceSets.test.get().output.classesDirs.asPath)
+    jvmArgs("-ea")
```
qwwdfsad, line 103: I'm a bit surprised it isn't added by default

build.gradle.kts
```diff
@@ -79,19 +63,12 @@ dependencies {
     implementation("org.ow2.asm:asm-tree:9.2")
     implementation("com.googlecode.java-diff-utils:diffutils:1.3.0")
     compileOnly("org.jetbrains.kotlin.multiplatform:org.jetbrains.kotlin.multiplatform.gradle.plugin:1.8.10")
-//    compileOnly("com.android.tools.build:gradle:${androidGradlePluginVersion}")
+    compileOnly("com.android.tools.build:gradle:${androidGradlePluginVersion}")
```
qwwdfsad, line 66: This one is worth keeping commented out until #108 is supported

build.gradle.kts
```diff
@@ -58,8 +43,7 @@ val createClasspathManifest = tasks.register("createClasspathManifest") {
         .withPropertyName("outputDir")
 
     doLast {
-        outputDir.mkdirs()
-        file(outputDir.resolve("plugin-classpath.txt")).writeText(testPluginRuntimeConfiguration.get().joinToString("\n"))
+        file(outputDir.resolve("plugin-classpath.txt")).writeText(testPluginRuntimeConfiguration.joinToString("\n"))
```
qwwdfsad, line 46: This one seems to be the really nasty hack from the earlier Gradle era where there were no proper test kits. Could you please check if it's still needed?

build.gradle.kts
```diff
@@ -22,29 +18,18 @@ sourceSets {
     }
 }
 
-sourceSets {
-    create("functionalTest") {
```
qwwdfsad, line 26: 👍 nice one with getting rid of it

qwwdfsad:
Thanks!

>Is the plugin still published as expected? All the artifacts should be the same.

It is not, as I fail to execute `./gradlew publishToMavenLocal` -- it now unconditionally requires PGP signature for local publication and testing. None of the developers [usually] have that, it would be nice to preserve the previous behaviour.

The corresponding error:
```
* What went wrong:
Execution failed for task ':signBinary-compatibility-validatorPluginMarkerMavenPublication'.
> Cannot perform signing task ':signBinary-compatibility-validatorPluginMarkerMavenPublication' because it has no configured signatory
```

build.gradle.kts
aSemy:
```suggestion
    //compileOnly("com.android.tools.build:gradle:${androidGradlePluginVersion}")
```

good point, I'm not sure why I uncommented it.

build.gradle.kts
aSemy: I'm happy to do so, but I thought it would be a bit of a larger change than just bumping the Gradle version. It would require updating the test code to remove the `addPluginTestRuntimeClasspath()` helper function.

Additionally, it seems like this hack is hardcoding the Kotlin version to be 1.8.10, even though the test datas specify Kotlin 1.3

https://github.com/Kotlin/binary-compatibility-validator/blob/5d08f51f5f5e10a9cbfdb72956018fe833d0621b/src/functionalTest/resources/examples/gradle/base/withPlugin.gradle.kts#L7

So when I remove the hack, the Config Cache checks fail. Not because BCV is incompatible with CC, but because Kotlin 1.3 is.

build.gradle.kts
aSemy: I think it's added by default too, but it was already set and I didn't want to change it in this PR.

build.gradle.kts
aSemy: thanks!

gradle.properties
```diff
@@ -2,6 +2,6 @@ version=0.13.0-SNAPSHOT
 group=org.jetbrains.kotlinx
 
 kotlinVersion=1.8.10
-pluginPublishVersion=0.10.1
+pluginPublishVersion=1.1.0
```
aSemy, line 5: Note to self: Bumping this version was required to be compatible with Gradle 8, but it probably caused the [signing requirement](https://github.com/Kotlin/binary-compatibility-validator/pull/122#issuecomment-1466521508).

build.gradle.kts
qwwdfsad: It would be totally okay, thanks!

qwwdfsad:
I confirmed that local publication works and that publishing works without breaking Sonatype invariants (e.g. it's releasable as is). Consumption of locally published (to mavenLocal) works as well, thanks!

It's already good to go, it's up to you whether you want to get rid of `plugin-classpath.txt` or proceed with other changes

aSemy:
> I confirmed that local publication works and that publishing works without breaking Sonatype invariants (e.g. it's releasable as is). Consumption of locally published (to mavenLocal) works as well, thanks!
> 
> It's already good to go, it's up to you whether you want to get rid of `plugin-classpath.txt` or proceed with other changes

Great! Thanks for checking. I think this can be merged. 

I tried replacing `plugin-classpath.txt` with the newer `withPluginClasspath()` Gradle TestKit helper, but it did not work. There were errors in the Kotlin Multiplatform test projects. What I think is happening is that the `testRuntime` classpath contains a version of the Kotlin plugin or library or Gradle API that clashes and overrides some JAR, the result being that BCV throws a 'class not found' error for the `KotlinProjectExtension` class. It was weird.

The solution would be to instead publish BCV to a local Maven directory (not necessarily Maven Local). I can do that in a separate PR.

So yes, go ahead and merge please!

Issue #88, opened 2022-06-21
Add an option to disable `allprojects` behavior.


JavierSegoviaCordoba:
`allprojects` is considered a bad practice in Gradle and it will break with future features like project isolation.

Should be great if one of these options was available before the final release:

- Add a flag to disable `allprojects` configuration
- Remove `allprojects` configuration and force it to apply to each project individually. To avoid duplicating code, the recommended practice nowadays is using convention plugins, so a simple example can be added to the readme.

Fleshgrinder:
Changing the plugin from a project plugin to a settings plugin would allow safe access to all projects without any hacks. It would also make it trivial to apply the plugin.

```kotlin
public abstract class SettingsPlugin : Plugin<Settings> {
    override fun apply(target: Settings) {
        target.gradle.apply {
            rootProject {
                // special handling for root project
            }

            beforeProject {
                // handling for ALL projects
            }
        }
    }
}
```

Obviously it is possible to introduce this in a backward compatible way:

```kotlin
public abstract class SettingsAndBuildPlugin : Plugin<Any> {
    override fun apply(target: Any) {
        when (target) {
            is Project -> {
                logger.error("{} should not be applied in build scripts anymore, apply it in the settings script of the build.", this::class.java.canonicalName)
                // apply
            }

            is Settings -> {
                // apply
            }

            else -> error("${this::class.java.canonicalName} cannot be applied to ${target::class.java.canonicalName}; it can only be applied to settings and build scripts.")
        }
    }
}
```

qwwdfsad:
The request itself is quite reasonable, though unfortunately, it's out of my scope right now.
I'm ready to accept a high-quality PR (with tests and docs) in a timely manner tho

aSemy:
I've done some experimenting, and I can pick this up. I'd first need to tidy up the build logic and doing some restructuring though.

* Update Gradle config to use buildSrc convention plugins
* Split up the current project into two projects 
   * the core (essentially everything the [`kotlinx.validation.api`](https://github.com/Kotlin/binary-compatibility-validator/tree/master/src/main/kotlin/api) package)
   * the Gradle plugin
* Finally, create a separate Settings plugin, and appropriate tests

@qwwdfsad I can make PRs for the above, if you approve? Would you like me to make separate issues for each step? The steps are also quite broad, and I will try and make the PRs smaller, but are there any steps that should be made more granular?

qwwdfsad:
[...]

PR #123, opened 2023-03-01
Initial Gradle convention plugins

aSemy:
- create [convention plugins](https://docs.gradle.org/current/userguide/custom_plugins.html#sec:precompiled_plugins) for common and java-specific config
- add Gradle [version catalog](https://docs.gradle.org/current/userguide/platforms.html#sub:central-declaration-of-dependencies)
- simplify some build config

Depends on #122 - and again, this PR should contain no functional changes to the BCV plugin or library.


qwwdfsad:
I've pushed `pr/122` branch -- if it works for you, this PR can point to it while we are working on #122 (so the changes are much easier to review). 

Or we can just wait until #122 is merged to avoid messing with branches

build.gradle.kts
```diff
@@ -58,21 +61,21 @@ configurations.implementation {
...
-    implementation("org.ow2.asm:asm:9.2")
-    implementation("org.ow2.asm:asm-tree:9.2")
-    implementation("com.googlecode.java-diff-utils:diffutils:1.3.0")
-    compileOnly("org.jetbrains.kotlin.multiplatform:org.jetbrains.kotlin.multiplatform.gradle.plugin:1.8.10")
-    compileOnly("com.android.tools.build:gradle:${androidGradlePluginVersion}")
+    implementation(libs.kotlinx.metadata)
+    implementation(libs.ow2.asm)
+    implementation(libs.ow2.asmTree)
+    implementation(libs.javaDiffUtils)
+    compileOnly(libs.gradlePlugin.android)
```
aSemy, line 68:
```suggestion
    //compileOnly(libs.gradlePlugin.android)
```

aSemy:
I think it would be best to wait for #122 before reviewing this one, but it is convenient to point it at a specific branch!

aSemy:
@qwwdfsad I think this PR was closed automatically by GitHub because the branch pr/122 was deleted. Could it be re-opened please?

qwwdfsad:
Hi, could you please elaborate whether you have any plans to push this change forward?

aSemy:
@qwwdfsad Ready for review! But could you update https://github.com/Kotlin/binary-compatibility-validator/tree/pr/122 to include the latest changes from master?

qwwdfsad:
I've abandoned `pr/122` the moment it was merged. I was assuming we are using `bcv-gradle-rework` branch for the upcoming work (https://github.com/Kotlin/binary-compatibility-validator/issues/88#issuecomment-1453557372).

Not a blocker anyway, will start reviewing, but could you please change the base branch and rebase on it?

buildSrc/src/main/kotlin/conventions/base.gradle.kts
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright 2016-2023 JetBrains s.r.o.
+ * Use of this source code is governed by the Apache 2.0 License that can be found in the LICENSE.txt file.
+ */
+package kotlinx.validation.build.conventions
```
qwwdfsad, line 5: Is it correct that these conventions will be published along with BCV and will be applicable to target's projects manually?

Could you please add an integration test for that?

buildSrc/src/main/kotlin/conventions/base.gradle.kts
```diff
@@ -0,0 +1,32 @@
...
+    project.group = rootProject.group
+}
+
+tasks.withType<AbstractArchiveTask>().configureEach {
+    // https://docs.gradle.org/current/userguide/working_with_files.html#sec:reproducible_archives
+    isPreserveFileTimestamps = false
+    isReproducibleFileOrder = true
+}
+
+tasks.withType<AbstractTestTask>().configureEach {
```
qwwdfsad, line 26: Could you please elaborate on the reasoning behind these two task types additional configurations?

qwwdfsad (commented):
So far, so good, thanks!

I have a few questions and a target branch comment, no comments about version catalog and I believe it's good to go

buildSrc/src/main/kotlin/conventions/base.gradle.kts
aSemy: These convention plugins are only available in BCV's own build scripts for building BCV, and they are not published anywhere. They won't be visible for users at all.

These are [pre-compiled script plugins](https://docs.gradle.org/current/userguide/custom_plugins.html#sec:precompiled_plugins) which Gradle will auto-magically compile into a plugin with a plugin ID (based on the file path). The plugins are compiled into a JAR, `buildSrc/build/libs/buildSrc.jar`, which ([because they're in buildSrc](https://docs.gradle.org/current/userguide/organizing_gradle_projects.html#sec:build_sources)) is automatically added to the buildscript classpath.

buildSrc/src/main/kotlin/conventions/base.gradle.kts
aSemy: They're common bits of config I've copied from my other projects.

Because these settings are set in a convention plugin, and plugins run their configuration a Gradle Project before any code in the `build.gradle.kts`, the settings can be overridden in specific tasks in some `subproject/build.gradle.kts` if required.

- `timeout.set(Duration.ofMinutes(60))` - this helps in rare cases where a test hangs, or doesn't exit properly. Without it, a test run might block a CI/CD pipeline, or continue running on someone's laptop without the owner noticing. When it's enabled Gradle will automatically quit the tests and mark them as failed, which is nice.

- `includeEmptyDirs = false` - I like this because when if `Sync` task that might filter out files, or move files around slightly, Gradle will still retain the original files. 

   Without this option you might Sync a file structure like this:
    
    ```
    .
    └── base/
        ├── alpha/
        │   └── file.txt
        └── beta/
            └── data.xml
    ```
    
    and if the Sync was filtered to only include `.txt` files, and dropped a single leading directory, then the destination would look like this.
    
    ```
    .
    └── base/
        ├── alpha/
        ├── beta/
        └── file.txt
    ```
    
    where there's two empty directories, `base/alpha` and `base/beta`. I've never needed to retain any empty directories, so I always set `includeEmptyDirs = false`. Rather than doing it in all tasks, it's nice just to do it once in a convention lpugin.

aSemy:
> I've abandoned `pr/122` the moment it was merged. I was assuming we are using `bcv-gradle-rework` branch for the upcoming work ([#88 (comment)](https://github.com/Kotlin/binary-compatibility-validator/issues/88#issuecomment-1453557372)).
> 
> Not a blocker anyway, will start reviewing, but could you please change the base branch and rebase on it?

Oops, I got the branches confused. I'll rebase.

buildSrc/src/main/kotlin/conventions/base.gradle.kts
qwwdfsad: Thanks!

qwwdfsad (approved):
Thanks!

PR #124, opened 2023-03-01
Split signature-generator and Gradle plugin into separate projects

aSemy:
As part of preparatory refactoring for #88, this PR splits up the signature-generator project and Gradle-plugin into two separate projects.

### Description

Because Gradle plugins must use Gradle's embedded version of Kotlin, splitting up the projects means the BCV Gradle plugin can be built using/targetting Gradle's embedded Kotlin, while the signature-generator can be built using another version of Kotlin.

It also means that the BCV Gradle plugin can use the `kotlin-dsl` plugin, which adds some Gradle specific configuration to help plugin development, but is not necessary in the signature-generator.

Splitting up the projects also helps prepare for using the Gradle Worker API, so when the BCV Gradle plugin runs the signature-generator it can specifically control the dependencies used, and potentially default to `kotlinx-metadata` dependency provided by KGP (if present).

### Summary of changes

- signatures-generator contains the JVM code, and is built using the standard KGP
- gradle-plugin contains only the Gradle Plugin code, and is built using Gradle's embedded-Kotlin
- Configuration between the two projects is shared using convention-plugins
- refactored Maven publication build config to use more Gradle idiomatic functionality, and because the signing plugin is flaky.
- I put them together in `./modules` dir. It's a personal convention that I use a lot of my projects, because it's nice that the built libraries are together under one directory and not mixed up with other build/gradle/IDE/git directories.

### Breaking changes

There should be no functional changes from this PR - the plugin and library should work exactly as before, however

* there are now two new artifacts instead of one
* The Gradle plugin ID is the same, so users who use the `plugins {}` DSL just need to bump the version.
* Because the signature-generator can only be published to Maven Central, and the default repository for plugins is `gradlePluginPortal()`, the BCV Gradle plugin won't be able to fetch the signature-generator by default

### TODO

- [x] **UPDATE: not an issue - GPP proxies Maven Central**
     Because the signature-generator can only be published to Maven Central, and the default repository for plugins is `gradlePluginPortal()`, the BCV Gradle plugin won't be able to fetch the signature-generator by default. There are a few ways around this:
   * update the docs, tell users to add `mavenCentral()` as a plugin repo
   * shadow signature-generator into BCV-Gradle https://plugins.gradle.org/docs/publish-plugin#shadowing
   * use the Worker API (maybe? I think this will fetch dependencies using the build-script repos)

- [ ] There are now two new artifacts, and the old one will continue to exist. Do you want to publish [Maven relocation information](https://docs.gradle.org/current/userguide/publishing_maven.html#publishing_maven:relocation)? 

### Dependencies

This PR depends on 

* #122 
* #123 
* #131


build.gradle.kts
aSemy, line 1: The configuration in this file was moved to buildSrc convention plugins, so that it could be shared between the two new subprojects as appropriate.

buildSrc/build.gradle.kts
```diff
@@ -3,42 +3,22 @@
...
-            else -> {
-                kotlin.setSrcDirs(emptyList<String>())
-                resources.setSrcDirs(emptyList<String>())
-            }
-        }
-        java.setSrcDirs(emptyList<String>())
-        groovy.setSrcDirs(emptyList<String>())
+    implementation("org.gradle.kotlin:gradle-kotlin-dsl-plugins:$expectedKotlinDslPluginsVersion") {
+        because("allows applying the `kotlin-dsl` / `embedded-kotlin` plugins in buildSrc convention plugins")
     }
```
aSemy, line 16: Sorry, I know you like removing the unnecessary src/main dirs, but in this case I thought it was extra configuration that didn't really bring much. Given how Gradle can be quite fragile, and how IntelliJ has tenuous support of buildSrc as it is, I thought it was better remove anything not strictly necessary.

qwwdfsad:
I've asked my colleague to help with the Gradle part, he mentioned the following potential caveats:

1) Compilation with embedded Kotlin might lead to LV/AV issues: e.g. with Gradle 8, embedded Kotlin immediately sets LV 1.8, rendering the plugin unusable with Gradle 6.9+. Probably can be fixed with explicit LV/AV for compilation

2) "the BCV Gradle plugin won’t be able to fetch the signature-generator by default". If I got this right, it's not an issues, because GPP proxies maven central as well

3) It's an open question (and it's unclear whether it's an issue at all) how to compile Gradle plugin against `signatures-generator` built with too new Kotlin version

aSemy:
> I've asked my colleague to help with the Gradle part, he mentioned the following potential caveats:
> 
> 1. Compilation with embedded Kotlin might lead to LV/AV issues: e.g. with Gradle 8, embedded Kotlin immediately sets LV 1.8, rendering the plugin unusable with Gradle 6.9+. Probably can be fixed with explicit LV/AV for compilation

Gradle Test Kit has been set to use Gradle 7.4.2 and the tests seems to work fine (although I don't really understand how or why...).

https://github.com/Kotlin/binary-compatibility-validator/blob/14bd77cb0406a2866bb6802eb073cd293bd47a06/src/functionalTest/kotlin/kotlinx/validation/api/TestDsl.kt#L34


> 2. "the BCV Gradle plugin won’t be able to fetch the signature-generator by default". If I got this right, it's not an issues, because GPP proxies maven central as well

Ah okay, interesting, thanks! And also, now that I think about it, it has to be the case. Lots of Gradle plugins depend on non-plugin dependencies, and they work fine. BCV depends on the Kotlin stdlib already, and that must come from Maven Central.

> 3. It's an open question (and it's unclear whether it's an issue at all) how to compile Gradle plugin against `signatures-generator` built with too new Kotlin version

Yes, good point. I expect that any problems would be picked up by the Gradle Test Kit test. Also, any further incompatibility problems would be reduced in the future by my plan to use a Gradle Worker to run `signatures-generator` in an isolated classpath.

qwwdfsad (changes requested):
Thank you, and sorry for the delay!

I gave it one more look, and it seems good in general
I have a few questions and an issue with non-conventional Gradle DSL and Maven Coordinates issue and I think it will be good to go

qwwdfsad:
Also, if it works for you, could you please update Kotlin to 1.9.0 alongside with this change?

PR #125, opened 2023-03-01
Apply `kotlin-dsl` plugin

aSemy:
Apply the [`kotlin-dsl` plugin](https://docs.gradle.org/current/userguide/kotlin_dsl.html#sec:kotlin-dsl_plugin) in the BCV Gradle project. 

This helps align the published plugin with the embedded Kotlin version. It also aids development, as it applies the SAM-with-receiver plugin, so the plugin code more closely resembles the syntax in `*.gradle.kts` files.

There should be no functional changes as a result of this PR.

Depends on 

* #122 
* #123 
* #124


modules/gradle-plugin/src/main/kotlin/KotlinApiBuildTask.kt
```diff
@@ -99,7 +99,7 @@ open class KotlinApiBuildTask @Inject constructor(
         val filteredSignatures = signatures
             .retainExplicitlyIncludedIfDeclared(publicPackages, publicClasses, publicMarkers)
             .filterOutNonPublic(ignoredPackages, ignoredClasses)
-            .filterOutAnnotated(nonPublicMarkers.map(::replaceDots).toSet())
+            .filterOutAnnotated(nonPublicMarkers.map { it.replace(".", "/") }.toSet())
```
aSemy, line 102: > Cannot access 'replaceDots': it is internal in 'kotlinx.validation.api'

`replaceDots()` is now in a separate subproject. Perhaps `replaceDots()` needs a `@RequiresOptIn` annotation? 

As a quick fix I copied the actual implementation.

aSemy:
I'm marking this as a draft - I'll come back to it once the required branches have been merged.

PR #126, opened 2023-03-01
Test util: Annotate file paths with `@Language("file-reference")`

aSemy:
Minor QOL improvement in the Test DSL. Annotating the file paths with `@Language("file-reference")` means the files can be opened in IntelliJ by cmd+clicking the file path.

Missing files are highlighted.

image

### Summary of changes

For this to work, 

* all paths need to be absolute from the root resources (all paths were already the full file path, I just needed to add the base `/`)
* use `java.getResouce()` instead of `classLoader.getResource()`.

### Impact

This is a test-only change and does not have any impact on the library.


qwwdfsad (approved):
Nice one! We definitely should leverage `@Language` more

PR #131, opened 2023-04-27
Apply BCV, commit API dump

aSemy:
Apply the BCV plugin.

I think we can all agree on why this is helpful :)

This PR depends on 

* https://github.com/Kotlin/binary-compatibility-validator/pull/123


qwwdfsad (approved):
LGTM, thanks. Will merge after #123 and will sort out ABI eventually, seems like it has a bit too many signatures in public

aSemy:
Splitting up the project also helps with organising the public API, the Gradle API dump is quite slim https://github.com/Kotlin/binary-compatibility-validator/pull/124/files#diff-28edfa00a73f55a3939d2d101d50f228ab00b993308b7a0fecffb34f7f73cfae

The signature-generator API dump is a little more chunky https://github.com/Kotlin/binary-compatibility-validator/pull/124/files#diff-dd8b72874dcb073e334d55e4b5d0b3850f5f3326a35afea5da50686224ac55f2

qwwdfsad (approved):
Thanks!

PR #136, opened 2023-05-04
ignoredClasses ignores classes that match a substring of the class name

### LINKED ISSUE #135: ignoredClasses ignores classes that match a substring of the class name
If I have two classes `com.example.Hello` and `com.example.HelloWorld`, and I add `com.example.HelloWorld` to my ignore classes list, apiDump will not generate dump information for `com.example.Hello`.

The issue seems to be the use of `startsWith` in 
https://github.com/Kotlin/binary-compatibility-validator/blob/master/src/main/kotlin/api/KotlinSignaturesLoading.kt#L246

`publicClasses.add("com.example.ktvalidator.Hello")` does not fix the issue.

bryanlogan:
If I have two classes com.example.Hello and com.example.HelloWorld, and I add com.example.HelloWorld to my ignore classes list, apiDump will not generate dump information for com.example.Hello.

Closes #135

### DIFF
--- a/src/main/kotlin/api/KotlinSignaturesLoading.kt
+++ b/src/main/kotlin/api/KotlinSignaturesLoading.kt
@@ -244,7 +244,7 @@ private fun ClassBinarySignature.isInPackages(packageNames: Collection<String>):
     packageNames.any { packageName -> name.startsWith(packageName) }
 
 private fun ClassBinarySignature.isInClasses(classNames: Collection<String>): Boolean =
-    classNames.any { className -> className.startsWith(name) }
+    classNames.any { className -> className.startsWith("$name/") }
 
 private fun JarFile.classEntries() = Sequence { entries().iterator() }.filter {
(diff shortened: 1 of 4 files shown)


src/main/kotlin/api/KotlinSignaturesLoading.kt
```diff
@@ -244,7 +244,7 @@ private fun ClassBinarySignature.isInPackages(packageNames: Collection<String>):
     packageNames.any { packageName -> name.startsWith(packageName) }
 
 private fun ClassBinarySignature.isInClasses(classNames: Collection<String>): Boolean =
-    classNames.any { className -> className.startsWith(name) }
+    classNames.any { className -> className.startsWith("$name/") }
```
fzhinkin, line 247: Now that the CBS's name is augmented with the `/`, `startsWith` could be replaced with an equality check, isn't it?

PR #143, opened 2023-06-13
Add possibility to check flavoured android modules

sebastienrouif:
Added the option to test flavoured android modules thanks to `testedFlavourName` parameter 
it would solve https://github.com/Kotlin/binary-compatibility-validator/issues/24

### DIFF
--- a/src/functionalTest/resources/examples/gradle/base/androidFlavouredKotlinLibrary.gradle.kts
+++ b/src/functionalTest/resources/examples/gradle/base/androidFlavouredKotlinLibrary.gradle.kts
@@ -2,6 +2,7 @@
  * Copyright 2016-2022 JetBrains s.r.o.
  * Use of this source code is governed by the Apache 2.0 License that can be found in the LICENSE.txt file.
  */
+import org.jetbrains.kotlin.gradle.plugin.*
 plugins {
     id("com.android.library")
     id("kotlin-android")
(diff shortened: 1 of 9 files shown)


src/main/kotlin/ApiValidationExtension.kt
```diff
@@ -64,4 +64,10 @@ open class ApiValidationExtension {
      * By default, only the `main` source set is checked.
      */
     public var additionalSourceSets: MutableSet<String> = HashSet()
+
+    /**
+     * In Android projects with multiple flavour,
```
maxtracking, line 69: typo: `flavours`

src/functionalTest/resources/examples/gradle/base/androidFlavouredKotlinLibrary.gradle.kts
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright 2016-2022 JetBrains s.r.o.
+ * Use of this source code is governed by the Apache 2.0 License that can be found in the LICENSE.txt file.
+ */
+import org.jetbrains.kotlin.gradle.plugin.*
```
maxtracking, line 5: nitpick: Should it follow the same rules as in other changed files for explicit imports? Or vice versa?

GokhanArik:
@qwwdfsad Is it possible to prioritise this PR?

NMGuner:
Does this take into account that some modules inside the project might lack flavors or the particular flavor set here? My suggestion would be to generate unique tasks for each flavor, similar to other gradle plugins compatible with flavors.

sebastienrouif:
Fair, I don't think I have checked that, let's try

shanshin:
As part of [the migration of separate BCV functionality to Kotlin Gradle Plugin](https://youtrack.jetbrains.com/issue/KT-71098/Migrate-Binary-Compatibility-Validator-to-Kotlin-Gradle-Plugin), the addition of new features to a separate plugin has been discontinued.

PR #144, opened 2023-06-14
Filter out $EntriesMappings class for Kotlin's 1.9 feature 'Enum entries'

### LINKED ISSUE #141: Filter out $EntriesMappings class for Kotlin's 1.9 feature 'Enum entries'
Kotlin 1.9 produces `public synthetic class` for each enum and `public static final synthetic field entries$0 Lkotlin/enums/EnumEntries;` in it. They should be filtered from public API dump.

fzhinkin:
Kotlin compiler generates synthetic class holding EnumEntries starting from 1.9. This change filters out such classes from the API dump.

Closes #141

### DIFF
--- a/src/main/kotlin/api/AsmMetadataLoading.kt
+++ b/src/main/kotlin/api/AsmMetadataLoading.kt
@@ -32,6 +32,7 @@ fun ClassNode.isEffectivelyPublic(classVisibility: ClassVisibility?) =
             && !isLocal()
             && !isWhenMappings()
             && !isSyntheticAnnotationClass()
+            && !isEnumEntriesMappings()
             && (classVisibility?.isPublic(isPublishedApi()) ?: true)
 
 
@@ -40,6 +41,7 @@ fun ClassNode.isLocal() = outerMethod != null
 fun ClassNode.isInner() = innerClassNode != null
 fun ClassNode.isWhenMappings() = isSynthetic(access) && name.endsWith("\$WhenMappings")
 fun ClassNode.isSyntheticAnnotationClass() = isSynthetic(access) && name.contains("\$annotationImpl\$")
+fun ClassNode.isEnumEntriesMappings() = isSynthetic(access) && name.endsWith("\$EntriesMappings")
 
 val ClassNode.effectiveAccess: Int get() = innerClassNode?.access ?: access
 val ClassNode.outerClassName: String? get() = innerClassNode?.outerName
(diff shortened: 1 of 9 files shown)


fzhinkin:
While the change is trivial, I don't see a way to test it easily.
I verified the behavior in a separate project, but I don't see how to pack it into a unit test that could be merged along with the change. 
Maybe someone has good ideas regarding it?

qwwdfsad:
Update Kotlin to 1.9.0-Beta will do the trick.
We do ensure compatibility via [`apiVersion`/`languageVersion`](https://kotlinlang.org/docs/compatibility-modes.html) and later even the concern with transitive dependencies will [cease to exist](https://youtrack.jetbrains.com/issue/KT-58999)

fzhinkin:
Updated to `1.8.20` instead, also removed exclusion of stdlib artifacts (it does nothing when the version of Kotlin from Gradle's classpath is lower, which is the case right now).

fzhinkin:
ping

src/main/kotlin/api/AsmMetadataLoading.kt
```diff
@@ -40,6 +41,7 @@ fun ClassNode.isLocal() = outerMethod != null
 fun ClassNode.isInner() = innerClassNode != null
 fun ClassNode.isWhenMappings() = isSynthetic(access) && name.endsWith("\$WhenMappings")
 fun ClassNode.isSyntheticAnnotationClass() = isSynthetic(access) && name.contains("\$annotationImpl\$")
+fun ClassNode.isEnumEntriesMappings() = isSynthetic(access) && name.endsWith("\$EntriesMappings")
```
qwwdfsad, line 44: There are currently no tests that check this behaviour.

You'll need `when` over old Kotlin enum (unlikely it's easy to get one in the robust manner) or any Java one in the code to force `$EntriesMapping` appearance

qwwdfsad (changes requested):

src/main/kotlin/api/AsmMetadataLoading.kt
qwwdfsad: No worries.

There is something suspicious happening:
 
1) IDE highlights `EnumClass.entries` as available only since API version 1.8 (red code, `[UNSUPPORTED_FEATURE]`) even though you clearly set it up
2) For clean builds, everything works as expected
3) For incremental builds (run all tests, change something in `entries.kt`), mappings start being generated even though `entries` member is present

qwwdfsad: I'll proceed with the issue further

qwwdfsad (approved):

Issue #92, opened 2022-07-11
Add support of semantic minor change


misha-dorokhin-fourthline:
I am building SDKs and follow [semantic versioning](https://semver.org/). It would be nice to have an option to detect the minor API update during `apiCheck` command. At this moment the added class would fail the `apiCheck` command.

But from SDK consumer point of view this change is compatible with the previous version because it does not require any effort in order to update to it.

Feature request:
- add an option to mark `minor` change as a non-breaking change. `apiCheck` should succeed
- print the added API to the console in this case

qwwdfsad:
Hi, 

It seems like what you are proposing is a higher level of abstraction than `binary-compatibility-validator` operates on. You can build such behaviour on top of BCV programmatic API, but BCV itself neither has notion of semantic versioning nor it should have:

* `apiCheck` is not aware of any versions at all and neither has ways to detect the version for which it checks API for
* It's not clear who is responsible for updating apiDump in that case. What prevents you from adding a new API in minor version `1` and removing it in minor version `2` if the dump is not updated?
*  Different libraries have different notions of compatibility. For example, Kotlin itself is not allowed to add any new non-experimental API in minor releases, while using this tool and we would like to keep it that way

PR #150, opened 2023-09-13
Add a config for apiCheck only failing on deletions and changes, and ignoring additions

mmollaverdi:
In places where we're using this plug-in, we've ended up ignoring the apiCheck failures at times, due to many failures being due to new APIs being added, which are not breaking changes. This new configuration makes it so that apiCheck only fails if the existing API lines have been `change`d or `delete`d, indicating either new arguments being added to an exisitng function/method or an existing function/method being deleted. Both of these situations are potential breaking changes and what we want to fail the bulid for.

### DIFF
--- a/src/main/kotlin/ApiValidationExtension.kt
+++ b/src/main/kotlin/ApiValidationExtension.kt
@@ -64,4 +64,9 @@ open class ApiValidationExtension {
      * By default, only the `main` source set is checked.
      */
     public var additionalSourceSets: MutableSet<String> = HashSet()
+
+    /**
+     * Only fail on API deletion and changes, and not on additions.
+     */
+    public var ignoreAdditions = false
 }
--- a/src/functionalTest/kotlin/kotlinx/validation/test/IgnoredAdditionsTests.kt
+++ b/src/functionalTest/kotlin/kotlinx/validation/test/IgnoredAdditionsTests.kt
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2016-2020 JetBrains s.r.o.
+ * Use of this source code is governed by the Apache 2.0 License that can be found in the LICENSE.txt file.
+ */
+
+package kotlinx.validation.test
+
+import kotlinx.validation.api.BaseKotlinGradleTest
+import kotlinx.validation.api.assertTaskSuccess
+import kotlinx.validation.api.buildGradleKts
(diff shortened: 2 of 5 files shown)


src/main/kotlin/ApiValidationExtension.kt
```diff
@@ -64,4 +64,9 @@ open class ApiValidationExtension {
      * By default, only the `main` source set is checked.
      */
     public var additionalSourceSets: MutableSet<String> = HashSet()
+
+    /**
+     * Only fail on API deletion and changes, and not on additions.
+     */
+    public var ignoreAdditions = false
```
mmollaverdi, line 71: Happy to take suggestions for this name, if it's confusing.
I'll update the documentation once we've settled on a name.

qwwdfsad (commented):
Thanks for the effort!

Unfortunately, we are not ready to accept your contribution, there was a request to support such behaviour, and we explicitly declined it, see https://github.com/Kotlin/binary-compatibility-validator/issues/70 (also: https://github.com/Kotlin/binary-compatibility-validator/issues/92).
The rationale is more or less straightforward:
* Not every removal or change is incompatible, and we would like to avoid semi-robust or obviously imprecise tools
* Additions might be binary compatible, but after they are released, their ignored absence is not; such configurations add one more failure mode to the tool that cannot be workarounded automatically because it is basically bound to the release cycle this tool is oblivious of


BCV can be used as a library though and one can emulate the desired behaviour in their project

mmollaverdi:
@qwwdfsad can I please clarify a few things

> Not every removal or change is incompatible, and we would like to avoid semi-robust or obviously imprecise tools

I deliberately avoided calling the new config `onlyFailOnIncompatibleChanges` or similar for this reasons. It's called `ignoreAdditions` which I think is explicit about what it does. The intention is to increase the signal-to-noise ratio, so that API check doesn't fail for additions which to my understanding are always compatible? Also note that this option is defaulted to false.

> Additions might be binary compatible, but after they are released, their ignored absence is not ...

There might be a misunderstanding here. The new `ignoreAddition` option, when enabled, only changes the behaviour of `apiCheck` so that it fails on removals and changes. It doesn't impact what `apiDump` does. The generated API dump file would still contain the new APIs. Does that address your concerns on "after they are released, their ignored absence is not"?

PR #153, opened 2023-10-18
Support KLib validation

fzhinkin:
Supported Klib ABI validation.

The validation is disabled by default and requires the following settings to be turned on:
```
apiValidation {
    klibValidationEnabled = true
}
```

KLib validation supports `ignoredPackages`, `ignoredClasses` and `nonPublicMarkers` options the same way classfiles validation does, but all `public*` options are unsupported and will be ignored.

As with the JVM API validation, ABI is dumped into a text file. There's a separate dump file for each native target, each located in a directory named after the target.
Here's an example of a file tree for kotlinx-io's core module:
```
kotlinx-io
    core
        api
            androidNativeArm32
                kotlinx-io-core.api
            androidNativeArm64
                kotlinx-io-core.api
...
            jvm
                kotlinx-io-core.api
...
            linuxArm64
                kotlinx-io-core.api
...
```

One of the disadvantages of such a scheme for storing dumps is that once something is changed in the common sourceset, api-files for all targets will reflect that change. It's an open question if some other scheme to represent the native ABI should be used.

### DIFF
--- a/src/functionalTest/resources/examples/gradle/configuration/ignoreSubclasses/ignore.gradle.kts
+++ b/src/functionalTest/resources/examples/gradle/configuration/ignoreSubclasses/ignore.gradle.kts
@@ -1,0 +1,9 @@
+/*
+ * Copyright 2016-2023 JetBrains s.r.o.
+ * Use of this source code is governed by the Apache 2.0 License that can be found in the LICENSE.txt file.
+ */
+
+configure<kotlinx.validation.ApiValidationExtension> {
+    ignoredClasses.add("subclasses.A.B")
+    //ignoredClasses.add("subclasses.A\$B")
+}
--- a/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
+++ b/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
@@ -78,9 +78,11 @@ class BinaryCompatibilityValidatorPlugin : Plugin<Project> {
             }
         }
 
+        // TODO: not sure how better handle it
         val dirConfig = jvmTargetCountProvider.map {
-            if (it == 1) DirConfig.COMMON else DirConfig.TARGET_DIR
+            /*if (it == 1) DirConfig.COMMON else */DirConfig.TARGET_DIR
         }
+        val nativeDirConfig = project.provider { DirConfig.NATIVE_TARGET_DIR }
 
         kotlin.targets.matching {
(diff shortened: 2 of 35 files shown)


src/functionalTest/resources/examples/gradle/configuration/ignoreSubclasses/ignore.gradle.kts
```diff
@@ -0,0 +1,9 @@
+/*
+ * Copyright 2016-2023 JetBrains s.r.o.
+ * Use of this source code is governed by the Apache 2.0 License that can be found in the LICENSE.txt file.
+ */
+
+configure<kotlinx.validation.ApiValidationExtension> {
+    ignoredClasses.add("subclasses.A.B")
+    //ignoredClasses.add("subclasses.A\$B")
```
fzhinkin, line 8: Subclass filtration is inconsistent between JVM API and Klib ABI dumps.
However, I believe that the problem should be addressed separately.

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -169,6 +181,9 @@ private enum class DirConfig {
      * `/api/jvm/project.api` and `/api/android/project.api`
      */
     TARGET_DIR,
+    // TODO
+    // Currently, has the same meaning as TARGET_DIR
+    NATIVE_TARGET_DIR
```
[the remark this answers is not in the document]
fzhinkin: To tackle the "dozens of identical files" issue, we can:
- create dumps for source sets sharing code identical among multiple targets (i.e. if there's only a common source set in KMP project, then we don't need to store dumps for each individual target), but there are multiple problems with such an approach.
- store files with a dump and an index file mapping targets to dumps; if dumps are identical for all targets, then there'll be only a single dump file and an index file will refer to it for all targets, but it'll be hard to track what has actually changed.
- store deduplicated dump-files and use symlinks to refer them for individual targets (not sure how well symlinks are going to work on Windows, though); cons here are the same as above.

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -103,6 +106,15 @@ class BinaryCompatibilityValidatorPlugin : Plugin<Project> {
                 }
             }
         }
+        kotlin.targets.matching { it.platformType == KotlinPlatformType.native }.all { target ->
```
fzhinkin, line 109: Wasm is missing here.

PR #154, opened 2023-10-23
Fix functional tests not utilizing the API_DIR const value

xtrm-en:
This PR just fixes the hardcoding of the `api` directory in the project's tests.


fzhinkin:
@xtrm-en thanks for the PR!

PR #162, opened 2023-12-15
Add all Companion class' annotations to corresponding Companion field.

### LINKED ISSUE #157: `FieldBinarySignature` which corresponds to `Companion` field should have companion annotations
In particular, a companion class, which is annotated with non-public marker, is excluded from the API, but the `Companion` field is still present.

Additionally, in IJ we analyze the presence of [`@ApiStatus.Experimental`](https://github.com/JetBrains/java-annotations/blob/b41de54f50e463211c71435c0fd86edeb9e3a152/common/src/main/java/org/jetbrains/annotations/ApiStatus.java#L54C25-L54C25), which is also absent from `FieldBinarySignature` of `Companion` field.

fzhinkin:
Fixes #157

### DIFF
--- a/src/main/kotlin/api/KotlinSignaturesLoading.kt
+++ b/src/main/kotlin/api/KotlinSignaturesLoading.kt
@@ -43,69 +43,25 @@ public fun Sequence<InputStream>.loadApiFromJvmClasses(visibilityFilter: (String
                 val supertypes = listOf(superName) - "java/lang/Object" + interfaces.sorted()
 
                 val fieldSignatures = fields
-                    .map {
-                        val annotationHolders =
-                            mVisibility?.members?.get(JvmFieldSignature(it.name, it.desc))?.propertyAnnotation
-                        val foundAnnotations = mutableListOf<AnnotationNode>()
-                        foundAnnotations.addAll(methods.annotationsFor(annotationHolders?.method))
-
-                        var companionClass: ClassNode? = null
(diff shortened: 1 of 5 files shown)


src/main/kotlin/api/KotlinSignaturesLoading.kt
```diff
@@ -46,23 +46,38 @@ public fun Sequence<InputStream>.loadApiFromJvmClasses(visibilityFilter: (String
                     .map {
                         val annotationHolders =
                             mVisibility?.members?.get(JvmFieldSignature(it.name, it.desc))?.propertyAnnotation
-                        val foundAnnotations = methods.annotationsFor(annotationHolders?.method)
-                        it.toFieldBinarySignature(foundAnnotations)
+                        val foundAnnotations = mutableListOf<AnnotationNode>()
```
qwwdfsad, line 49: `loadApiFromJvmClasses` method is kind of hard to follow with all the incremental additions.

If you don't mind, would be nice to start extracting parts of it into separate bodies (e.g. this `map` body) with meaningful names

qwwdfsad (approved):

src/main/kotlin/api/KotlinSignaturesLoading.kt
qwwdfsad: Also, it will allow using typealias with a meaningful name instead of a pair

qwwdfsad (approved):
Nice!

PR #167, opened 2024-01-12
Do not depend on sources and javadocs twice during the publication pr…

qwwdfsad:
…ocess

* 'java { withXJar }' already provides a dependency

Fixes 'Invalid publication 'maven': multiple artifacts with the identical extension and classifier ('jar', 'javadoc')' failure during publication process


fzhinkin (approved):
Thanks for addressing the issue!

PR #168, opened 2024-01-12
Enable strict explicit API mode

qwwdfsad:
Rationale:

We know have quite a lot of accidentally public entities while both promoting and using BCV as a standalone JAR dependency. Apart from that, we have quite an unfortunate package name 'api' that might imply all these methods are part of public API (when, in fact, it's all related to API validation).

It would be nice to explicitly confine our visibilities and be more deliberate about that

### DIFF
--- a/src/main/kotlin/KotlinApiCompareTask.kt
+++ b/src/main/kotlin/KotlinApiCompareTask.kt
@@ -15,7 +15,7 @@ import org.gradle.api.file.*
 import org.gradle.api.model.ObjectFactory
 import org.gradle.api.tasks.*
 
-open class KotlinApiCompareTask @Inject constructor(private val objects: ObjectFactory): DefaultTask() {
+public open class KotlinApiCompareTask @Inject constructor(private val objects: ObjectFactory): DefaultTask() {
 
     /*
      * Nullability and optionality is a workaround for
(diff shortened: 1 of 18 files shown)


src/main/kotlin/KotlinApiCompareTask.kt
```diff
@@ -26,14 +26,14 @@ open class KotlinApiCompareTask @Inject constructor(private val objects: ObjectF
...
     @InputDirectory
     @PathSensitive(PathSensitivity.RELATIVE)
-    var projectApiDir: File? = null
+    public var projectApiDir: File? = null
 
     // Used for diagnostic error message when projectApiDir doesn't exist
     @Input
     @Optional
-    var nonExistingProjectApiDir: String? = null
+    public var nonExistingProjectApiDir: String? = null
```
fzhinkin, line 34: Does anyone actually use it?

fzhinkin: In fact, I'm interested in how you decided which subset of gradle tasks' API should remain public.

qwwdfsad: I made everything I could `internal`/`private`. These pieces are required by Gradle machinery.

`@Input`-like things have to be public, it's checked by Gradle plugin validation (`:validatePlugin` that is run on each `check`)

fzhinkin (approved):

PR #170, opened 2024-01-16
Configure output directory for dumps

### LINKED ISSUE #127: Support output folder configuration
Our mono-repo uses a hierarchy of projects where the API definitions are kept in a sub-folder named `api` under each project. When I run `apiDump`, those checked in folders all get overridden because by default the plugin writes the `api` folder. Is there any way to customize the output folder name?

fzhinkin:
Added an option to configure a path to a directory where dumps are stored:

``` 
apiValidation {
   apiDumpDirectory = "a/b/c/d/e"
}
```

By default, that path remains `api`.

It's an error to specify a path that leads outside the project directory.

Fixes #127

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,11 @@ apiValidation {
      * Flag to programmatically disable compatibility validator
      */
     validationDisabled = true
+
+    /**
+     * A path to a subdirectory inside the project root directory, where dumps should be stored.
+     */
+    apiDumpDirectory = "api"
 }
 ```
 
@@ -123,6 +128,11 @@ apiValidation {
      * Flag to programmatically disable compatibility validator
      */
     validationDisabled = false
+
+    /**
+     * A path to a subdirectory inside the project root directory, where dumps should be stored.
+     */
+    apiDumpDirectory = "aux/validation"
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ apiValidation {
     validationDisabled = true
 
     /**
-     * A path to a subdirectory inside the project root directory, where dumps should be stored.
+     * A path to a subdirectory inside the project root directory where dumps should be stored.
      */
     apiDumpDirectory = "api"
 }
@@ -130,7 +130,7 @@ apiValidation {
     validationDisabled = false
 
     /**
-     * A path to a subdirectory inside the project root directory, where dumps should be stored.
+     * A path to a subdirectory inside the project root directory where dumps should be stored.
      */
     apiDumpDirectory = "aux/validation"
 }
--- a/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
+++ b/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
@@ -138,7 +138,17 @@ private class TargetConfig constructor(
     val targetName: String? = null,
     private val dirConfig: Provider<DirConfig>? = null,
 ) {
-    private val API_DIR_PROVIDER = project.provider { extension.apiDumpDirectory }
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -89,6 +89,11 @@ apiValidation {
      * Flag to programmatically disable compatibility validator
      */
     validationDisabled = true
+
+    /**
+     * A path to a subdirectory inside the project root directory, where dumps should be stored.
```
qwwdfsad, line 94:
```suggestion
     * A path to a subdirectory inside the project root directory where dumps should be stored.
```

README.md
```diff
@@ -123,6 +128,11 @@ apiValidation {
      * Flag to programmatically disable compatibility validator
      */
     validationDisabled = false
+
+    /**
+     * A path to a subdirectory inside the project root directory, where dumps should be stored.
```
qwwdfsad, line 133:
```suggestion
     * A path to a subdirectory inside the project root directory where dumps should be stored.
```

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -130,17 +128,27 @@ public class BinaryCompatibilityValidatorPlugin : Plugin<Project> {
         project: Project,
         extension: ApiValidationExtension
     ) = configurePlugin("kotlin", project, extension) {
-        project.configureApiTasks(extension, TargetConfig(project))
+        project.configureApiTasks(extension, TargetConfig(project, extension))
     }
 }
 
 private class TargetConfig constructor(
```
qwwdfsad, line 135: Maybe something like the following would be more straightforward?

```
private class TargetConfig(
    project: Project,
    val targetName: String? = null,
    dirConfig: Provider<DirConfig>? = null,
) {

    val apiDir = dirConfig?.map { dirConfig ->
        when (dirConfig) {
            DirConfig.COMMON -> API_DIR
            else -> "$API_DIR/$targetName"
        }
    } ?: project.provider { API_DIR }
    

    fun apiTaskName(suffix: String) = when (targetName) {
        null, "" -> "api$suffix"
        else     -> "${targetName}Api$suffix"
    }
    
}
```

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -150,10 +158,10 @@ private class TargetConfig constructor(
     val apiDir
```
qwwdfsad, line 158: Could you please move the property above the `fun` block?

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -130,17 +128,27 @@ public class BinaryCompatibilityValidatorPlugin : Plugin<Project> {
...
     private val dirConfig: Provider<DirConfig>? = null,
 ) {
+    private val apiDirProvider = project.provider {
+        val dir = extension.apiDumpDirectory
 
-    private val API_DIR_PROVIDER = project.provider { API_DIR }
+        val root = project.layout.projectDirectory.asFile.toPath().toAbsolutePath().normalize()
+        val resolvedDir = root.resolve(dir).normalize()
+        if (!resolvedDir.startsWith(root)) {
+            throw IllegalArgumentException("apiDumpDirectory should be inside the project directory")
```
qwwdfsad, line 147: Please consider enriching the message with the contextual information about expected and actual paths

qwwdfsad (approved):
👍

qwwdfsad:
(I removed the obsolete comment because I was looking at the wrong branch in my IDE)

PR #175, opened 2024-01-25
Support Java package-level annotations

### LINKED ISSUE #156: Support package-level non-public marker
Instead of specifying a list of packages, it should be possible to specify a marker, for example [`@ApiStatus.Internal`](https://github.com/JetBrains/java-annotations/blob/b41de54f50e463211c71435c0fd86edeb9e3a152/common/src/main/java/org/jetbrains/annotations/ApiStatus.java#L73C25-L73C25)

fzhinkin:
Support extraction of packages annotated with particular annotations. These packages could be then be retained as public packages or filtered out as ignored packages.

Closes #156

### DIFF
--- a/src/main/kotlin/api/KotlinSignaturesLoading.kt
+++ b/src/main/kotlin/api/KotlinSignaturesLoading.kt
@@ -218,12 +218,23 @@ private fun List<ClassBinarySignature>.filterOutNotAnnotated(
     }
 }
 
+/**
+ * Extracts name of packages annotated by one of the [targetAnnotations].
+ * If there are no such packages, returns an empty list.
+ *
+ * Package is checked for being annotated by looking at classes with `package-info` name
+ * ([see JSL 7.4.1](https://docs.oracle.com/javase/specs/jls/se21/html/jls-7.html#jls-7.4)
+ * for details about `package-info`).
+ */
 @ExternalApi
 public fun List<ClassBinarySignature>.extractAnnotatedPackages(targetAnnotations: Set<String>): List<String> {
     if (targetAnnotations.isEmpty()) return emptyList()
 
     return filter {
         it.name.endsWith("/package-info")
+    }.filter {
+        // package-info classes are private synthetic abstract interfaces since 2005 (JDK-6232928).
+        it.access.isInterface && it.access.isSynthetic && it.access.isAbstract
     }.filter {
         it.annotations.any {
             ann -> targetAnnotations.any { ann.refersToName(it) }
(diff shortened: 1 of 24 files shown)


src/main/kotlin/api/KotlinSignaturesLoading.kt
```diff
@@ -218,6 +218,19 @@ private fun List<ClassBinarySignature>.filterOutNotAnnotated(
     }
 }
 
+@ExternalApi
+public fun List<ClassBinarySignature>.extractAnnotatedPackages(targetAnnotations: Set<String>): List<String> {
+    if (targetAnnotations.isEmpty()) return emptyList()
+
+    return filter {
+        it.name.endsWith("/package-info")
```
qwwdfsad, line 226: This code looks like something in the form of "@Ann object \`package-info`" might slip in, but haven't checked it

src/main/kotlin/api/KotlinSignaturesLoading.kt
```diff
@@ -218,6 +218,19 @@ private fun List<ClassBinarySignature>.filterOutNotAnnotated(
     }
 }
 
+@ExternalApi
```
qwwdfsad, line 221: Could you please elaborate on why it's public API?

src/test/kotlin/tests/CasesPublicAPITest.kt
```diff
@@ -59,11 +66,23 @@ class CasesPublicAPITest {
 
     private fun snapshotAPIAndCompare(testClassRelativePath: String, nonPublicMarkers: Set<String> = emptySet()) {
         val testClassPaths = baseClassPaths.map { it.resolve(testClassRelativePath) }
-        val testClasses = testClassPaths.flatMap { it.listFiles().orEmpty().asIterable() }
+        val testClasses = testClassPaths.flatMap {
+            if (!it.exists()) return@flatMap emptyList()
+            val allFiles = mutableListOf<File>()
+            Files.walkFileTree(it.toPath(), object : SimpleFileVisitor<Path>() {
```
qwwdfsad, line 72: We have a nice API for that:
```
it.toPath().visitFileTree { 
    onVisitFile { file, attributes ->
        allFiles.add(file.toFile())
    }
}
```

Also, worth filing KTIJ for the corresponding quickfix

src/main/kotlin/api/KotlinSignaturesLoading.kt
fzhinkin: It's an API facilitating markers gathering and if someone is using the BCV as an API, then they probably will need this as well.

qwwdfsad: Maybe it's worth adding a bit of KDoc then

src/main/kotlin/api/KotlinSignaturesLoading.kt
fzhinkin: I narrowed the set of matching `package-info` declarations to what javac should actually emit (abstract synthetic interfaces), thanks for pointing to it!

qwwdfsad (approved):
Nice!

Issue #149, opened 2023-09-01
Validate KLIB ABI compatibility on Multiplatform


fzhinkin:
There's an API to verify KLIB's binary compatibility that'll be released soon: 
https://youtrack.jetbrains.com/Issue/KT-54402
https://github.com/JetBrains/kotlin/blob/master/compiler/util-klib-abi/ReadMe.md

BCV should use it to support KLIBs validation in addition to JVM bytecode validation.

qwwdfsad:
Things that are worth keeping in mind (or maybe split into a few subtasks): dependency on Kotlin compiler:

* We can hardcode the dependency for the first version, but then we'll have to update it each Kotlin release
* We can later provide a plugin extension that takes the dependency of the same version as KGP applied to the project

ddolovov:
I would recommend to use `v2` of binary signatures (with mangled names included into the rendered signature text) vs `v1` (with only hashes).

fzhinkin:
> What is the naming scheme/migration path for non-JVM dumps, and how are they different (or the same?) across platform-specific, JS IR and WASM source-sets

I see a few options here:
1) use unified naming for both jvm and non-jvm target: `api/<target>/<projectName>.api`
-    pros: we already use such naming if a project contains multiple jvm-targets;
- cons: for an existing multiplatform project it means that an old api-file for a jvm-target need to be moved to `jvm` subdirectory.
2) for jvm targets don't do anything, for native targets use naming scheme mentioned above
-   pros: much pleasant migration path as there's no need to update jvm-specific dumps
- cons: there will be a dependency between the jvm'a apiDump task and native tasks, need to look carefully how to deal with it
3) dump klibs to their own `abi`-directory: `abi/<target>/<projectName>.abi`
- pros: easier to implement (compared to (2))
- cons: adds some complexity by introducing an additional directory for dumps: its unclear why KLIB ABI dumps deserved a different top-level directory
   

It's also unclear if there's anything we should do about identical KLIB ABI dumps across different native targets.
For example, for `kotlinx-io-core`, dumps for `linuxX64`, `linuxArm64`, `mingwX64` and all androidNative*-targets are identical, as well as all dumps for apple-targets are identical.
I don't really like the idea of storing identical files within the project and I don't like the idea of updating all these dumps once an API changed even more. But at the same time it's unclear if having identical ABI dumps for almost all targets is a common case (I would guess that it is) and how to group these dumps (assuming that they can diverge in the future).
[...]

PR #181, opened 2024-01-30
Update Kotlin version to 1.9.22

fzhinkin:
As part of KLib support (#149) we need to bump the Kotlin version as the required API is only available starting from the 1.9.20.

### DIFF
--- a/src/functionalTest/kotlin/kotlinx/validation/api/TestDsl.kt
+++ b/src/functionalTest/kotlin/kotlinx/validation/api/TestDsl.kt
@@ -13,7 +13,7 @@ import org.intellij.lang.annotations.Language
 public val API_DIR: String = ApiValidationExtension().apiDumpDirectory
 
 internal fun BaseKotlinGradleTest.test(
-    gradleVersion: String = "7.4.2",
+    gradleVersion: String = "8.5",
     injectPluginClasspath: Boolean = true,
     fn: BaseKotlinScope.() -> Unit
 ): GradleRunner {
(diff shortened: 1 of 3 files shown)


src/functionalTest/resources/examples/gradle/base/multiplatformWithJvmTargets.gradle.kts
```diff
@@ -21,6 +21,9 @@ kotlin {
             testRuns["test"].executionTask.configure {
                 useJUnit()
             }
+            attributes {
```
fzhinkin, line 24: It seems like `functionalTests` always use the same Kotlin version as the projects, so after updating to 1.9.22 we need to declare build variants here. Otherwise, the configuration won't succeed.

src/functionalTest/kotlin/kotlinx/validation/api/TestDsl.kt
```diff
@@ -13,7 +13,7 @@ import org.intellij.lang.annotations.Language
 public val API_DIR: String = ApiValidationExtension().apiDumpDirectory
 
 internal fun BaseKotlinGradleTest.test(
-    gradleVersion: String = "7.4.2",
+    gradleVersion: String = "8.5",
```
qwwdfsad, line 16: Do we change the minimal supported Gradle version for BCV? 
If so, worth noting it in README

qwwdfsad (approved):

src/functionalTest/kotlin/kotlinx/validation/api/TestDsl.kt
fzhinkin: I'll double check it. But here the version was updated only to make tests work with 1.9.22 (as I mentioned elsewhere, that version leaks into functional tests and I don't know what to do with that).

fzhinkin: I don't think we ever mentioned it explicitly, but the minimum supported Gradle version is 6.0 and after updating to Kotlin 1.9.22 the plugin still works with it.

qwwdfsad: Thanks!

PR #182, opened 2024-01-31
Added minimum required Gradle version

fzhinkin:


qwwdfsad (approved):

README.md
```diff
@@ -21,6 +22,10 @@ The tool allows dumping binary API of a JVM part of a Kotlin library that is pub
   * [Class member changes](#class-member-changes)
 * [Building locally](#building-the-project-locally)
 
+## Requirements
+
+Binary compatibility validator plugin requires Gradle `6.0` or newer.
```
qwwdfsad, line 27: The only risk I see here is that we don't test it on a regular basis. But this problem is widespread across all our Gradle plugins

PR #183, opened 2024-01-31
Klib support

fzhinkin:
This PR introduces KLib ABI validation support in the BCV plugin.

The brief overview of the design and decisions made is in the corresponding design doc: https://github.com/Kotlin/binary-compatibility-validator/blob/50ebf8b00492d754732e47453b4d8d0351f33baa/docs/design/KLibSupport.md

Providing a programmatic API similar to the `kotlinx.validation.api` API for working with JVM ABI is planned, but this PR does not include it.

Besides the design doc mentioned above, the main entry points for reviewing are https://github.com/Kotlin/binary-compatibility-validator/blob/cac5669a1d403e90b2a33516239f1c470af32819/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt#L370 where all the Gradle tasks are configured and https://github.com/Kotlin/binary-compatibility-validator/blob/cac5669a1d403e90b2a33516239f1c470af32819/src/main/kotlin/klib/KlibAbiDumpFileMerger.kt#L77 where the current dump merging logic reside.


docs/design/KLibSupport.md
```diff
@@ -0,0 +1,184 @@
...
+So, the class `BuildConfig` have the same targets as the whole file, but `linuxSpecific`-function
+is presented only on `linux`-targets (`// Targets: [linux]`).
+
+The next two lines declare target aliases (`// Alias: <name> => [<targets>]`). There are only two
+aliases (`linux` and `androidNative`), because it seems excessive to declare aliases for groups
+consisting of only a single target (like `mingw`) as well as declaring aliases for targets
+than will never be explicitly presented in a file (like an alias for all the targets, or all native targets). 
+We also can't add aliases for targets that does not exist, so there are no aliases like `macos` or `apple` 
+(in this example).
+However, all other aliases are generated even if it is not explicitly mentioned (like `androidNative`) to
```
fzhinkin, line 93: @ilya-g we discussed it previously and decided to stick with "declare all possible aliases, even unused", but amount of wording required to describe why some aliases are mentioned whilst other aren't makes me feel that something here is not right.

fzhinkin: *so it might be worth reconsidering which aliases should be printed

src/main/kotlin/klib/KlibAbiDumpFileMerger.kt
```diff
@@ -0,0 +1,614 @@
...
+            "Expected a file staring with \"$expectedHeader\", but the file stats with \"$headerStart\""
+        }
+    }
+
+    private fun LinesProvider.parseDeclaration(
+        depth: Int,
+        parent: DeclarationContainer,
+        allTargets: Set<Target>,
+        isMergedFile: Boolean,
+        aliases: Map<String, Set<Target>>
```
fzhinkin, line 252: Feels like all these 100500-parameters deserved a single ParsingContext object (and childrenCache should also be its part and should not be included into parsed declarations), but I think that could be fixed separately.

fzhinkin: > and childrenCache should also be its part and should not be included into parsed declarations

It seems like it worked incorrectly anyway, so I got rid of it.

sandwwraith (commented):
I didn't look at the implementation itself, but I decided to post comments now, because they were pending for too long

src/main/kotlin/ApiValidationExtension.kt
```diff
@@ -71,4 +71,52 @@ public open class ApiValidationExtension {
...
+@ExperimentalBCVApi
+public open class KlibValidationSettings {
+    /**
+     * Enables KLib ABI validation checks.
+     */
+    public var enabled: Boolean = false
+    /**
+     * Specify which version of signature KLib ABI dump should contain.
+     */
+    public var signatureVersion: Int = 2
```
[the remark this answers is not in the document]
fzhinkin: We can, that's for sure. But when the default/latest version will change, old dumps will be invalid. So there should be a way to specify a particular version.
But here we should probably use some `latest-version-available` alias instead of the concrete version `2`.

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -11,10 +11,17 @@ import org.gradle.api.provider.*
...
+import org.jetbrains.kotlin.konan.target.HostManager
+import org.jetbrains.kotlin.library.abi.ExperimentalLibraryAbiReader
+import org.jetbrains.kotlin.library.abi.LibraryAbiReader
 import java.io.*
+import kotlin.text.split
 
 public class BinaryCompatibilityValidatorPlugin : Plugin<Project> {
 
+    @ExperimentalLibraryAbiReader
+    @ExperimentalBCVApi
```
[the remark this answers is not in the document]
fzhinkin: No actual pros, we can simply opt-in both annotations for the whole class (or even the file).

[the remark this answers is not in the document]
sandwwraith: I'd prefer opt-in then, and leave experimentality only on actual configuration flag

src/main/kotlin/ApiValidationExtension.kt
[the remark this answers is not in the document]
sandwwraith: I don't get it. If the klib version changes, it means `.klib.api` file format (not actual signatures) would also change? That looks undesirable. Probably we should have some intermediate parser layer that would provide stable output.

 Also, if klib version changes to 3, but user has `signatureVersion = 2` it means that we parse klib v3 but output should look like 2. Which brings us again to some middle-layer transformation.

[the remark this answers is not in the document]
fzhinkin: A klib may have multiple signature versions. For example, each declaration may have both v1 and v2 signatures.
The signature is a part of a dump we're producing, so when it changes, the dump also change.

> Probably we should have some intermediate parser layer that would provide stable output.

Not sure how we can do that. Signatures are part of a declaration as a linker looks up for symbols by their signatures. And there may not be a direct mapping between different signature versions and a declaration they belongs to.

> Also, if klib version changes to 3, but user has signatureVersion = 2 it means that we parse klib v3 but output should look like 2. Which brings us again to some middle-layer transformation.

We parse the same klib, but instead of dumping v2 signature, will dump v3 signature.

A klib's dump with v1 signature:
https://github.com/Kotlin/binary-compatibility-validator/blob/f5fdacf4d2e9548144f522aa2924742ac4e473f5/src/functionalTest/resources/examples/classes/TopLevelDeclarations.klib.v1.dump
The dump of the same klib using v2 signature:
https://github.com/Kotlin/binary-compatibility-validator/blob/f5fdacf4d2e9548144f522aa2924742ac4e473f5/src/functionalTest/resources/examples/classes/TopLevelDeclarations.klib.dump

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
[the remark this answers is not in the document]
fzhinkin: Done

src/main/kotlin/ApiValidationExtension.kt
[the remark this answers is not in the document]
sandwwraith: So, v3 actually can't link with v2 because they're incompatible? But newer klibs will have both 3 and 2 sets for backwards compatibility?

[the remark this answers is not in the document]
fzhinkin: > So, v3 actually can't link with v2 because they're incompatible?

Yes, the linker won't be able to lookup symbols.

> But newer klibs will have both 3 and 2 sets for backwards compatibility?

There's no v3 on the horizon right now, so it's hard to tell.

src/main/kotlin/ApiValidationExtension.kt
```diff
@@ -71,4 +71,52 @@ public open class ApiValidationExtension {
...
+    public var strictValidation: Boolean = false
+    /**
+     * For declarations available only on some targets, replace a comment listing all targets
+     * with a short alias.
+     *
+     * Group aliases are based on [the default hierarchy template](https://kotlinlang.org/docs/multiplatform-hierarchy.html#see-the-full-hierarchy-template)
+     * and enabled by default.
+     *
+     * Grouping will not be applied if a project has a target with a custom name that clashes
+     * with one of the group names.
```
ilya-g, line 119: We have discussed that we should use target names as they are written in the manifest of klib, this way setting custom names would have no effect on dumps.

src/main/kotlin/ApiValidationExtension.kt
```diff
@@ -71,4 +71,52 @@ public open class ApiValidationExtension {
...
+     */
+    public var signatureVersion: Int = 2
+    /**
+     * Fail validation if some build targets are not supported by the host compiler.
+     * By default, ABI dumped only for supported files will be validated. This option makes validation behavior
+     * stricter and treats having unsupported targets as an error.
+     */
+    public var strictValidation: Boolean = false
+    /**
+     * For declarations available only on some targets, replace a comment listing all targets
```
ilya-g, line 112:
```suggestion
     * For declarations available only on some targets, replace a comment listing these targets
```

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -11,8 +11,14 @@ import org.gradle.api.provider.*
 import org.gradle.api.tasks.*
 import org.jetbrains.kotlin.gradle.dsl.*
 import org.jetbrains.kotlin.gradle.plugin.*
+import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
+import org.jetbrains.kotlin.konan.target.HostManager
+import org.jetbrains.kotlin.library.abi.ExperimentalLibraryAbiReader
+import org.jetbrains.kotlin.library.abi.LibraryAbiReader
 import java.io.*
+import kotlin.text.split
```
ilya-g, line 19: minor: likely redundant

src/main/kotlin/ApiValidationExtension.kt
```diff
@@ -71,4 +71,52 @@ public open class ApiValidationExtension {
...
+ * Settings affecting KLib ABI validation.
+ */
+@ExperimentalBCVApi
+public open class KlibValidationSettings {
+    /**
+     * Enables KLib ABI validation checks.
+     */
+    public var enabled: Boolean = false
+    /**
+     * Specify which version of signature KLib ABI dump should contain.
```
ilya-g, line 102: "Specifies"

src/main/kotlin/ApiValidationExtension.kt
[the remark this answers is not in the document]
ilya-g: The docs here should guide user when to rely on the default value and when to set it to a fixed value.

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -299,3 +329,331 @@ private inline fun <reified T : Task> Project.task(
     name: String,
     noinline configuration: T.() -> Unit,
 ): TaskProvider<T> = tasks.register(name, T::class.java, Action(configuration))
+
+private const val BANNED_TARGETS_PROPERTY_NAME = "binary.compatibility.validator.klib.targets.blacklist.for.testing"
```
ilya-g, line 333: Consider `binary.compatibility.validator.klib.targets.disabled.for.testing`

src/main/kotlin/ApiValidationExtension.kt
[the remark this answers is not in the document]
fzhinkin: Set the `null` as a default, mentioned that a user should not care about the option by default.

src/main/kotlin/ApiValidationExtension.kt
```diff
@@ -71,4 +71,52 @@ public open class ApiValidationExtension {
...
+     */
+    public var signatureVersion: Int = 2
+    /**
+     * Fail validation if some build targets are not supported by the host compiler.
+     * By default, ABI dumped only for supported files will be validated. This option makes validation behavior
+     * stricter and treats having unsupported targets as an error.
+     */
+    public var strictValidation: Boolean = false
```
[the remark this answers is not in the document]
fzhinkin: Configuration cache seems to work fine with primitive values. But if there's a case where Properties are better, I can change it (note that a parent class, `ApiValidationExtension` don't use Property too).

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
fzhinkin: Removed, thanks!

src/main/kotlin/ApiValidationExtension.kt
fzhinkin: Done, thanks!

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
fzhinkin: Renamed

src/main/kotlin/ApiValidationExtension.kt
fzhinkin: After a discussion, we agreed to include only aliases that could actually appear in a dump.

fzhinkin: After a discussion we agreed on the following scheme:
- dump will include user-facing target names (target names that were configured in a build script);
- however,  an underlying target will be used for grouping purposes;
- if a target has a custom name clashing with pre-defined groups, grouping will be disabled for such a dump.

For example, here
```
kotlin {
    macosArm64("linuxArm64") // haha, take that, plugin!
    linuxX64()
}
```
the dump will include "linuxArm64" and "linuxX64" as target names, but there target won't be grouped together into a "linux" group as the underlying target for a target named "linuxArm64" is "macos_arm64", not a linux.

docs/design/KLibSupport.md
```diff
@@ -0,0 +1,212 @@
...
+
+However, a klib's manifest contains all the information required to find an actual underlying target (but it does
+not contain a custom name, though). And the same info could be included in a textual klib dump.
+
+To overcome the issue, it is proposed to represent target name as a tuple consisting of "visible" name
+and a canonical underlying target name: `name.canonicalName`.
+
+For the example mentioned above, target such fully qualified target names are:
+- `macos.macosArm64` for `macosArm64("macos")`;
+- `linux.linuxArm64` for `linuxArm64("linux")`;
```
ilya-g, line 197: I thought we agreed on `linuxArm64.linux` for this situation

fzhinkin: I put the configured name first as something that is then clarified with an underlying target name. But I don't mind swapping these components. We can briefly discuss it today.

src/main/kotlin/klib/KlibAbiDumpFileMerger.kt
```diff
@@ -0,0 +1,733 @@
...
+        if (nextLine != null) {
+            val res = nextLine!!
+            nextLine = null
+            return res
+        }
+        return lines.next()
+    }
+}
+
+private const val MERGED_DUMP_FILE_HEADER = "// KLib ABI Dump"
```
ilya-g, line 42: Also "Klib" here. Or just "klib"

src/main/kotlin/klib/KlibAbiDumpFileMerger.kt
```diff
@@ -0,0 +1,733 @@
...
+    private val topLevelDeclaration: DeclarationContainer = DeclarationContainer("")
+
+    /**
+     * All targets for which this dump contains declarations.
+     */
+    internal val targets: Set<KlibTarget> = targetsMut
+
+    internal fun merge(file: File, configurableTargetName: String? = null) {
+        require(file.exists()) { "File does not exist: $file" }
+        Files.lines(file.toPath()).use {
```
ilya-g, line 96: Sad that we cannot use `file.toPath().useLines { }` yet. If we up gradle requirements for the plugin in future, this can be left as a TODO

src/main/kotlin/klib/KlibAbiDumpFileMerger.kt
```diff
@@ -0,0 +1,733 @@
...
+        topLevelDeclaration.targets.addAll(bcvTargets)
+
+        // All declarations belonging to the same scope have equal indentation.
+        // Nested declarations have higher indentation.
+        // By tracking the indentation, we can decide if the line should be added into the current container,
+        // to its parent container (i.e., the line represents sibling declaration) or the current declaration ended,
+        // and we must pop one or several declarations out of the parsing stack.
+        var currentContainer = topLevelDeclaration
+        var depth = -1
+        val targetsStack = Stack<Set<KlibTarget>>().apply { push(bcvTargets) }
```
ilya-g, line 144: Minor: usually `MutableList` works better as a stack, than `Stack`.

src/main/kotlin/klib/KlibAbiDumpFileMerger.kt
```diff
@@ -0,0 +1,733 @@
...
+
+package kotlinx.validation.klib
+
+import kotlinx.validation.api.klib.KlibTarget
+import java.io.File
+import java.nio.file.Files
+import java.util.*
+import kotlin.Comparator
+
+private class LinesProvider(private val lines: Iterator<String>) : Iterator<String> {
```
ilya-g, line 14: Makes sense to name it `PeekingLineIterator`

src/main/kotlin/klib/KlibAbiDumpFileMerger.kt
```diff
@@ -0,0 +1,733 @@
...
+        }
+
+        if (platformString == "WASM") {
+            // Currently, there's no way to distinguish Wasm targets without explicitly specifying a target name
+            check(configurableTargetName != null) { "targetName has to be specified for a Wasm target" }
+            return setOf(KlibTarget(configurableTargetName))
+        }
+        if (platformString != "NATIVE") {
+            return if (configurableTargetName == null) {
+                setOf(KlibTarget(platformString.toLowerCase()))
```
ilya-g, line 295: While `lowercase()` is not available, use `toLowerCase` with an explicit Locale (and not the default one).

src/main/kotlin/klib/KlibAbiDumpFileMerger.kt
```diff
@@ -0,0 +1,733 @@
...
+        targetsMut.clear()
+        targetsMut.addAll(targets)
+
+        topLevelDeclaration.overrideTargets(targets)
+    }
+
+    internal fun visit(action: (DeclarationContainer) -> Unit) {
+        topLevelDeclaration.children.forEach {
+            action(it.value)
+        }
```
ilya-g, line 484: Minor:
```suggestion
        topLevelDeclaration.children.values.forEach(action)
```

src/main/kotlin/klib/KlibAbiDumpFileMerger.kt
```diff
@@ -0,0 +1,733 @@
...
+
+// TODO: optimize
+private object DeclarationsComparator : Comparator<DeclarationContainer> {
+    override fun compare(c0: DeclarationContainer, c1: DeclarationContainer): Int {
+        return if (c0.targets == c1.targets) {
+            c0.text.compareTo(c1.text)
+        } else {
+            if (c0.targets.size == c1.targets.size) {
+                val c0targets = c0.targets.asSequence().map { it.toString() }.sorted().iterator()
+                val c1targets = c1.targets.asSequence().map { it.toString() }.sorted().iterator()
```
ilya-g, line 635: For small collection sizes, the laziness usually doesn't pull its weight.
Also `sorted` materializes all elements before sorting.

Switching to eager operations will result in an additional list instance, or even in none, if you mapTo(mutableListOf()) and sort it in place - `apply { sort() }`

src/main/kotlin/klib/TargetHierarchy.kt
```diff
@@ -0,0 +1,150 @@
...
+    data class NodeClosure(val node: Node, val allLeafs: Set<String>)
+
+    internal val hierarchyIndex: Map<String, NodeClosure>
+
+    private val hierarchy = Node(
+        "all",
+        Node("js"),
+        Node("wasmJs"),
+        Node("wasmWasi"),
+        Node("wasm32"),
```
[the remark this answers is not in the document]
fzhinkin: I intentionally supported it as DSL methods responsible for its configuration are still there. I double checked, and KMP won't let you build them, so, I guess, it could be removed from here too (along with MIPS targets).

src/main/kotlin/klib/TargetHierarchy.kt
```diff
@@ -0,0 +1,150 @@
...
+
+        init {
+            childrenNodes.forEach {
+                it.parent = this
+            }
+        }
+
+        fun <T> visit(visitor: (Node) -> T): T {
+           return visitor(this)
+        }
```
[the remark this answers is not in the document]
fzhinkin: Its one and only use is collecting leaf node, so it's better to introduce `collectLeafs(hierarchy)`, yeah.

src/main/kotlin/klib/KlibAbiDumpFileMerger.kt
ilya-g: > **and not the default one**

With the default locale, you can get the same problems as with the overload without a locale, e.g. Turkish i problem

src/main/kotlin/klib/KlibAbiDumpFileMerger.kt
```diff
@@ -0,0 +1,733 @@
...
+            }
+        }
+    }
+
+    fun formatDeclarationTargets(targets: Set<KlibTarget>): String {
+        val targetsMut = targets.toMutableSet()
+        val resultingTargets = mutableListOf<String>()
+        for (alias in aliases) {
+            if (targetsMut == alias.targets) {
+                targetsMut.removeAll(alias.targets)
```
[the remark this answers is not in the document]
ilya-g: By the current logic, does it mean that if a declaration has all native and for example one Js targets, its native target list won't be replaced with an alias?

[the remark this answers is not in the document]
fzhinkin: That's correct (but it doesn't feel right). I'll revert that logic back.

Note that, after that, it'll replace all native targets with "native" only if there is a declaration available only in all native targets:
```
// Targets: [js, linuxX64, ... <all other native>]
public fun nativeAndJs(): Int = 1
// Targets: [linuxX64, ... <all other native> ]
public fun nativeOnly(): Int = 1
```
will be transformed into
```
// Alias native: ...
// Targets: [js, native]
public fun nativeAndJs(): Int = 1
// Targets: [native]
public fun nativeOnly(): Int = 1
```
but the alias will not be introduced without `nativeOnly` function.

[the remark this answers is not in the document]
fzhinkin: Perhaps, an alias should be introduced only when a declaration has the exact list of corresponding targets, but also if it includes the whole list (like with the `nativeAndJs` above).

@ilya-g WHYT?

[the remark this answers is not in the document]
fzhinkin: Here are example of how it behaves now and how it could/should: https://gist.github.com/fzhinkin/4ac1535d7e6428079978067e08630fc6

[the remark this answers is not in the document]
ilya-g: Yes, the latest proposal dump seems more convenient and understandable.

As I imagine, the logic should be something like following:
- we start we the largest group in the hierarchy, and if it's entirely contained in the set of declaration targets, we replace these targets with the group alias
- then we repeat with the smaller groups down the hierarchy.

src/main/kotlin/api/klib/KlibDumpFilters.kt
```diff
@@ -0,0 +1,199 @@
...
+import org.jetbrains.kotlin.gradle.utils.`is`
+import org.jetbrains.kotlin.library.abi.*
+import java.io.File
+import java.io.FileNotFoundException
+
+/**
+ * Filters affecting how the klib ABI will be represented in a dump.
+ */
+@ExperimentalBCVApi
+public class KLibDumpFilters internal constructor(
```
ilya-g, line 18: Klib

src/main/kotlin/api/klib/KlibDump.kt
```diff
@@ -0,0 +1,276 @@
...
+ * ```kotlin
+ * val dump = KlibDump.from(File("/path/to/dump.klib.api"))
+ * ```
+ *
+ * **Merging multiple dumps into a new merged dump**
+ * ```kotlin
+ * val klibs = listOf(File("/path/to/library-linuxX64.klib"), File("/path/to/library-linuxArm64.klib"), ...)
+ * val mergedDump = KlibDump()
+ * klibs.forEach { mergedDump.mergeFromKlib(it) }
+ * File("/path/to/merged.klib.api").bufferedWriter().use { mergedDump.saveTo(it) }
```
ilya-g, line 33: It turns out quite asymmetric that dumps can be loaded only from files, but can be saved only to Appendables.

src/main/kotlin/api/klib/KlibDump.kt
```diff
@@ -0,0 +1,276 @@
...
+        require(intersection.isEmpty()) {
+            "Cannot merge dump as this and other dumps share some targets: $intersection"
+        }
+        merger.merge(other.merger)
+    }
+
+    /**
+     * Remove all declarations that do not belong to specified targets and remove these targets from the dump.
+     *
+     * All targets not stored within a dump will be ignored.
```
ilya-g, line 106:
```suggestion
     * All targets in the [targets] collection not contained within this dump will be ignored.
```

src/main/kotlin/api/klib/KlibDump.kt
```diff
@@ -0,0 +1,276 @@
...
+    public fun merge(other: KlibDump) {
+        val intersection = targets.intersect(other.targets)
+        require(intersection.isEmpty()) {
+            "Cannot merge dump as this and other dumps share some targets: $intersection"
+        }
+        merger.merge(other.merger)
+    }
+
+    /**
+     * Remove all declarations that do not belong to specified targets and remove these targets from the dump.
```
ilya-g, line 104:
```suggestion
     * Removes all declarations that do not belong to specified targets and removes these targets from the dump.
```

src/main/kotlin/api/klib/KlibAbiDumpFileMerger.kt
```diff
@@ -0,0 +1,741 @@
...
+    private fun merge(lines: PeekingLineIterator, configurableTargetName: String?) {
+        require(lines.peek() != null) { "File is empty" }
+        val isMergedFile = lines.determineFileType()
+
+        val aliases = mutableMapOf<String, Set<KlibTarget>>()
+        val bcvTargets = mutableSetOf<KlibTarget>()
+        if (isMergedFile) {
+            lines.next() // skip the heading line
+            bcvTargets.addAll(lines.parseTargets(configurableTargetName))
+            check(bcvTargets.size == 1 || configurableTargetName == null) {
```
ilya-g, line 115: Can this check fail if the `isMergedFile` is false?
For example, there can potentially be klibs with the multiple native targets in their manifest, how it would look like in the non-merged dump?

fzhinkin: For non-merged klibs, there is a similar check inside `extractTargets` function that will fail if there are multiple native targets and configurableTargetName is not null

src/main/kotlin/api/klib/KlibDumpFilters.kt
fzhinkin: sed betrayed me 😭

src/main/kotlin/api/klib/KlibDump.kt
fzhinkin: Added a `saveTo(file: File)` extension.
As of parsing from char sequences, not sure if it's needed outside our tests.

ilya-g (approved):
Forgot to submit the last portion of comments

PR #186, opened 2024-02-05
Add incremental validation

diareuse:
Validation is modified when using `useIncrementalValidation=true` to not fail when encountering additions, which are known to be backwards compatible. Instead spits out warnings to allow developers commiting the file early with the changes they have made.

The use case is to pair it with CI-run :check task and commit updated API after merge - considering all steps are automated. In cases where the API is not compatible, the task fails similarly to `useIncrementalValidation=false` or simply _the default_.

It's acknowledged that using `useIncrementalValidation=true` might cause issues when building upon a feature (or PR) in which cases the .api file might be accidentally prematurely commited to the VCS. In which case the developer might prefer leaving the option in the default state.

To the naked eye it might be apparent that checking only removals (leading `-`) is naive, but since the api diff add two distinct lines for changes (one leading `-`, other `+`), this naive approach proves working for all considered use-cases.

---

Change also includes ignored test, which ensures that `apiDump` is run automatically. I was not able, unfortunately, to get this feature working due to time constraints and not wanting to introduce additional Gradle tasks to automate this task. Maintainers are advised to either delete the test or implement the feature at their own pace. 🙂


fzhinkin:
@diareuse could you please [file an issue](https://github.com/Kotlin/binary-compatibility-validator/issues/new) describing the problem you're solving, examples of how and when the new validation mode should be used, as well as listing alternatives and shortcomings of the proposed approach?
Besides being a good reference point, an issue also seems more suitable for discussing a new feature, its implications, and alternatives.

diareuse:
Will do

diareuse:
@fzhinkin can we have a followup on this? :) Did you get a second opinion on this feature perhaps?

PR #192, opened 2024-03-14
Eliminate trailing double newline in dump

### LINKED ISSUE #133: Improve handling of trailing whitespace in API dump files
Currently, any API dump that BCV produces contains two trailing whitespaces lines, `\n\n`.

I have enabled 'strip trailing whitespace' in IntelliJ https://stackoverflow.com/questions/20679842/remove-trailing-whitespace-on-save-in-intellij-idea-12

When I open a .api file, which might automatically happen during a commit dialog, IntelliJ will replace the two trailing `\n\n` with a single `\n`. This is annoying, because then BCV will fail the test.

Other editors might also have problems with different types of behaviour that are difficult to predict or control.

Could BCV be changed to ignore any difference in trailing whitespace?

JakeWharton:
Closes #133

### DIFF
--- a/src/functionalTest/kotlin/kotlinx/validation/api/TestDsl.kt
+++ b/src/functionalTest/kotlin/kotlinx/validation/api/TestDsl.kt
@@ -27,8 +27,10 @@ internal fun BaseKotlinGradleTest.test(
                 createNewFile()
             }
 
-        scope.files.forEach {
+        scope.files.forEachIndexed { index, it ->
             val fileContent = readFileList(it)
+
+            if (index > 0) fileWriteTo.appendText("\n")
             fileWriteTo.appendText(fileContent)
         }
     }
--- a/src/test/kotlin/tests/utils.kt
+++ b/src/test/kotlin/tests/utils.kt
@@ -18,14 +18,13 @@ fun List<ClassBinarySignature>.dumpAndCompareWith(to: File) {
         to.bufferedWriter().use { dump(to = it) }
         fail("Expected data file did not exist. Generating: $to")
     } else {
-        val actual = dump(to = StringBuilder())
+        val actual = dump(to = StringBuilder()).toString()
         assertEqualsToFile(to, actual)
     }
 }
 
-private fun assertEqualsToFile(expectedFile: File, actual: CharSequence) {
-    val actualText = actual.trimTrailingWhitespacesAndAddNewlineAtEOF()
-    val expectedText = expectedFile.readText().trimTrailingWhitespacesAndAddNewlineAtEOF()
+private fun assertEqualsToFile(expectedFile: File, actualText: String) {
+    val expectedText = expectedFile.readText()
 
     if (OVERWRITE_EXPECTED_OUTPUT && expectedText != actualText) {
         expectedFile.writeText(actualText)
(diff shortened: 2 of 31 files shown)


src/functionalTest/kotlin/kotlinx/validation/api/TestDsl.kt
```diff
@@ -27,8 +27,10 @@ internal fun BaseKotlinGradleTest.test(
                 createNewFile()
             }
 
-        scope.files.forEach {
+        scope.files.forEachIndexed { index, it ->
```
JakeWharton, line 30: The fact that you can add multiple dump files and have them magically joined is weird. It's only used in one test. That test should specify a combined file and this functionality should be eliminated.

Can do in a follow-up if you agree.

src/test/kotlin/tests/utils.kt
```diff
@@ -34,8 +33,3 @@ private fun assertEqualsToFile(expectedFile: File, actual: CharSequence) {
 
     assertEquals(expectedText, actualText, "Actual data differs from file content: ${expectedFile.name}\nTo overwrite the expected API rerun with -Doverwrite.output=true parameter\n")
 }
-
-private fun CharSequence.trimTrailingWhitespacesAndAddNewlineAtEOF(): String =
```
JakeWharton, line 38: Another testing oddity. This fundamentally undermines the ability to validate the textual representation of the dump and prevented actually asserting against the behavior change in this PR.

The presence or absence of trailing whitespace on a line and the presence or absence of trailing newlines are part of the output and should be validated as such.

JakeWharton:
Sure, no rush. Thanks. I'll do a rebase in the mean time later today.

src/functionalTest/kotlin/kotlinx/validation/api/TestDsl.kt
fzhinkin: It's not only about dumps, and I personally find it quite convenient when it comes to constructing Gradle build scripts, like here: https://github.com/Kotlin/binary-compatibility-validator/blob/218728f09404b5e4330678fb9101191954b3e182/src/functionalTest/kotlin/kotlinx/validation/test/InputJarTest.kt#L18

fzhinkin:
I'm not 100% sure if a double newline at the end of a dump was intentional (@ilya-g correct me if it was), but changing it right now will annoy a lot of users as their dumps will no longer pass validation.

So, at least `KotlinApiCompareTask` has to be update to ignore trailing newlines.
I would also add a `dump` overload accepting a parameter controlling trailing newlines and preserve the current behavior for the old overload, so that existing `dump` users (like Kotlin stdlib) won't experience any changes in a dump format:
```
public fun <T : Appendable> List<ClassBinarySignature>.dump(to: T): T = dumpTo(to, true)
public fun <T : Appendable> List<ClassBinarySignature>.dump(to: T, addTrailingNewline: Boolean): T { .. }
```

JakeWharton:
The format has changed before in minor version bumps. Since the next version is a minor version bump, perhaps we can instead warn if the double trailing newline is present. And then remove that for 0.16 where we only accept the new format.

fzhinkin:
> The format has changed before in minor version bumps.

That's true, but it was usually associated with improvements in public ABI handling, not just the format itself.

> Since the next version is a minor version bump, perhaps we can instead warn if the double trailing newline is present. And then remove that for 0.16 where we only accept the new format.

The fix is about cleaning up the mess BCV made previously itself, not something users deliberately made themselves. So I don't think we should bother users if there's an option to slightly change `KotlinApiCompareTask`'s behavior and make everything unnoticeable (for those who don't update the dump, ofc).

PR #200, opened 2024-03-21
Correctly handle compilations without sources

### LINKED ISSUE #199: Klib validation failing on CI but not locally
```
* What went wrong:	
Execution failed for task ':redwood-layout-modifiers:klibApiMerge'.	
> java.io.FileNotFoundException: File does not exist: /Users/runner/work/redwood/redwood/redwood-layout-modifiers/build/api/js/redwood-layout-modifiers.klib.api	
```
- Project: https://github.com/cashapp/redwood
- PR: https://github.com/cashapp/redwood/pull/1884
- SHA: 5dc8c3377b3c3b7760416dcc5308033090213ca3

Attempt 1
 - Action log: https://github.com/cashapp/redwood/actions/runs/8360157932/job/22885183135?pr=1884
 - Build scan: https://gradle.com/s/p5ha2lzccuwiq

Attempt 2
 - Action log: https://github.com/cashapp/redwood/actions/runs/8360157932/job/22891922690?pr=1884
 - Build scan: https://gradle.com/s/ehehhmdxrrf5a

Attempt 3
 - Action log: https://github.com/cashapp/redwood/actions/runs/8360157932/job/22893510501?pr=1884
 - Build scan: https://gradle.com/s/tiijq75sgyt4c

I have not been able to reproduce this locally on my Mac. I tried executing the same command (after a `clean`) with `--max-workers=3` to match the core count of the CI machine but it always succeeds (5 attempts).

I also don't know why it's always this specific module and the JS target that fails. I apply it to ~40 modules in this build, most of which are multiplatform containing JS targets. Very weird.

It's possible the problem can be identified just from looking at the code of the plugin, but I don't have time to do that right now. I can try investigating more later tonight or tomorrow, but figured I'd file the issue first.

fzhinkin:
BCV's Klib-validation part does not handle compilations without any sources properly, as a result tasks manipulating dumps might expect a dump for a target that doesn't have any compiled klibs.

Such targets/compilations occur when, for example, a project consists only of test sources.

Verified the fix locally using both projects mentioned in the original bug report. 

Fixes #199

### DIFF
--- a/src/functionalTest/kotlin/kotlinx/validation/test/KlibVerificationTests.kt
+++ b/src/functionalTest/kotlin/kotlinx/validation/test/KlibVerificationTests.kt
@@ -18,6 +18,7 @@ import org.junit.Test
 import java.io.File
 import java.nio.file.Files
 import java.nio.file.Paths
+import kotlin.test.assertFalse
 import kotlin.test.assertTrue
 
 internal const val BANNED_TARGETS_PROPERTY_NAME = "binary.compatibility.validator.klib.targets.disabled.for.testing"
@@ -632,5 +633,40 @@ internal class KlibVerificationTests : BaseKotlinGradleTest() {
                         "androidNativeX86, linuxArm64.linux, linuxX64, mingwX64]"
             )
         }
+    }
+
+    @Test
+    fun `apiDump should not fail for empty project`() {
+        val runner = test {
+            baseProjectSetting()
+            addToSrcSet("/examples/classes/AnotherBuildConfig.kt", sourceSet = "commonTest")
+            runApiDump()
+        }
+
+        runner.build().apply {
+            assertTaskSkipped(":klibApiDump")
+        }
+        assertFalse(runner.projectDir.resolve("api").exists())
+    }
+
+    @Test
+    fun `apiDump should not fail if there is only one target`() {
+        val runner = test {
+            baseProjectSetting()
+            addToSrcSet("/examples/classes/AnotherBuildConfig.kt", sourceSet = "commonTest")
+            addToSrcSet("/examples/classes/AnotherBuildConfig.kt", sourceSet = "linuxX64Main")
+            runApiDump()
+        }
+        checkKlibDump(runner.build(), "/examples/classes/AnotherBuildConfig.klib.linuxX64Only.dump")
+    }
+
+    @Test
+    fun `apiCheck should not fail for empty project`() {
+        val runner = test {
+            baseProjectSetting()
+            addToSrcSet("/examples/classes/AnotherBuildConfig.kt", sourceSet = "commonTest")
+            runApiCheck()
+        }
+        runner.build()
     }
 }
--- a/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
+++ b/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
@@ -10,8 +10,8 @@ import kotlinx.validation.api.klib.konanTargetNameMapping
 import org.gradle.api.*
 import org.gradle.api.plugins.*
 import org.gradle.api.provider.*
+import org.gradle.api.specs.Spec
 import org.gradle.api.tasks.*
-import org.jetbrains.kotlin.cli.jvm.main
 import org.jetbrains.kotlin.gradle.dsl.*
 import org.jetbrains.kotlin.gradle.plugin.*
 import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
@@ -473,7 +473,13 @@ private class KlibValidationPipelineBuilder(
                 "into a single merged KLib ABI dump"
         dumpFileName = klibDumpFileName
         mergedFile = klibMergeDir.resolve(klibDumpFileName)
-        compilableTargets = project.compilableTargets()
+        val compilableTargets = project.compilableTargets()
+        filterTargets(Spec<String> {
+                name -> compilableTargets.get()
+            .map(KlibTarget::parse)
+            .map { it.configurableName }
+            .contains(name)
+        })
         onlyIf {
             compilableTargets.get().isNotEmpty()
         }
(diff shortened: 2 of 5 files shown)


src/main/kotlin/KotlinKlibMergeAbiTask.kt
```diff
@@ -45,16 +46,25 @@ internal abstract class KotlinKlibMergeAbiTask : DefaultTask() {
...
+    }
+
     @OptIn(ExperimentalBCVApi::class)
     @TaskAction
     internal fun merge() {
+        val filter = targetPredicate
         KlibDump().apply {
             targetToFile.forEach { (targetName, dumpDir) ->
-                merge(dumpDir.resolve(dumpFileName), targetName)
+                if (filter.isSatisfiedBy(targetName)) {
```
fzhinkin, line 65: Perhaps, instead of all this fuss with filtration, it would be better to simply check if a dump file exists and skip merging it, if it's not.

I don't see a way to filter out not-compilable targets (targets without sources) when configuring the merge task, so the best thing we can do here is check which targets were actually compiled during the task execution.

Filtration here serves as a safety net to catch issues when we expected a dump (`isSatisfiedBy` returns true), but for some reason it was not generated (file does not exist), but the only reason why it could happen is a misconfiguration of some sort.

WDYT?

fzhinkin: Assuming the task will be made public, it's probably better to remove the filter and merge only existing dumps (i.e. check if a file exists).

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -579,6 +593,30 @@ private class KlibValidationPipelineBuilder(
...
+        return project.provider {
+            project.kotlinMultiplatform.targets.matching { it.emitsKlib }
+                .asSequence()
+                .filter { it.mainCompilationOrNull?.hasAnySources() == true }
+                .map { KlibTarget(extractUnderlyingTarget(it), it.targetName).toString() }
+                .toSet()
+        }
+    }
+
+    private fun Project.isTargetWithConfigurableNameCompilable(): Spec<String> {
```
shanshin, line 607: I can't see any usage of this

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -579,6 +593,30 @@ private class KlibValidationPipelineBuilder(
         }
     }
 
+    // Targets having some sources to compile
+    private fun Project.compilableTargets(): Provider<Set<String>> {
```
shanshin, line 597: If we use this function only inside `onlyIf` block, so we may rename it and return `Provider<Boolean>`

src/functionalTest/kotlin/kotlinx/validation/test/KlibVerificationTests.kt
```diff
@@ -633,4 +634,39 @@ internal class KlibVerificationTests : BaseKotlinGradleTest() {
...
+            baseProjectSetting()
+            addToSrcSet("/examples/classes/AnotherBuildConfig.kt", sourceSet = "commonTest")
+            addToSrcSet("/examples/classes/AnotherBuildConfig.kt", sourceSet = "linuxX64Main")
+            runApiDump()
+        }
+        checkKlibDump(runner.build(), "/examples/classes/AnotherBuildConfig.klib.linuxX64Only.dump")
+    }
+
+    @Test
+    fun `apiCheck should not fail for empty project`() {
```
shanshin, line 664: What about test to check merging of two targets, but one of them have no sources?

fzhinkin: There's `apiDump should not fail if there is only one target` below, where there are multiple targets, but only one of them has sources.

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -676,10 +711,14 @@ private fun extractUnderlyingTarget(target: KotlinTarget): String {
...
+private val KotlinTarget.mainCompilationOrNull: KotlinCompilation<KotlinCommonOptions>?
+    get() = compilations.firstOrNull { it.name == KotlinCompilation.MAIN_COMPILATION_NAME }
 
 private val Project.jvmDumpFileName: String
     get() = "$name.api"
 private val Project.klibDumpFileName: String
     get() = "$name.klib.api"
+
+private fun KotlinCompilation<KotlinCommonOptions>.hasAnySources(): Boolean = allKotlinSourceSets.any {
+    it.kotlin.srcDirs.any(File::exists)
```
JakeWharton, line 723: This seems like it will break our usage in Redwood where we have no source dirs but we add a directory-generating task to the source set. The directory is not created until the task is executed, but you seem to be checking at configuration-time.

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
fzhinkin: Removed, thanks!

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
fzhinkin: I didn't check it for all Redwood subprojects, but for `:redwood-layout-modifiers` generated source directories are already there at the time of this call.
Note that I only extracted this logic into a separate method, it was already used for the JVM-part of the plugin before.

shanshin: @fzhinkin, what about using `onlyIf` + `Provider<Boolean>` instead of `isEnabled`?

fzhinkin: Sure, why not

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -579,6 +584,14 @@ private class KlibValidationPipelineBuilder(
         }
     }
 
+    // Returns a predicate that checks if there are any compilable targets
+    private fun Project.compilableTargetsPredicate(): Provider<Boolean> {
```
shanshin, line 588: `hasAnyCompilableTarget`?

shanshin (approved):

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
fzhinkin: I would expect such a function to return a boolean, not a predicate.

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
fzhinkin: Well, turned out that it may break JVM tasks which relied on dependencies being disabled eagerly. :(

fzhinkin: https://github.com/Kotlin/binary-compatibility-validator/pull/212

PR #202, opened 2024-03-25
Correctly handle dumps starting with Targets-list

### LINKED ISSUE #201: Klib validation failing with unexpected header line error
Running `apiCheck` results in the following error:
```
Unexpected header line: // Targets: [apple]
```

The error is referring to [this](https://github.com/rickclephas/KMP-NativeCoroutines/blob/571fad4af395d7383dc2c01981ac1f3f7f69dba7/kmp-nativecoroutines-core/api/kmp-nativecoroutines-core.klib.api#L10) line. Adding an empty line between `// Library unique name:` and `// Targets:` seems to solve the issue (but is overridden by `apiDump`).

fzhinkin:
Fixes #201


fzhinkin:
Additionally verified the fix for the project mentioned in the original bug report.

sandwwraith (approved):

PR #204, opened 2024-03-26
Make Klib-related Gradle tasks public

### LINKED ISSUE #203: Make KLib validation related tasks public
Currently, all tasks required for klib validation are internal.

We should make these tasks public (and cleanup the API along the way), so those who can't directly apply our plugin could incorporate them into custom plugins without writing identical tasks using programmatic API.

fzhinkin:
What was done:
- cleaned up klib-related tasks API
- rewrote code gluing up tasks together to utilize Gradle properties for proper dependency tracking
- made tasks public
- made some programmatic API classes used as tasks inputs serializable

Fixed #203

### DIFF
--- a/src/main/kotlin/-Utils.kt
+++ b/src/main/kotlin/-Utils.kt
@@ -10,6 +10,7 @@ import kotlinx.validation.api.klib.konanTargetNameMapping
 import org.gradle.api.file.RegularFileProperty
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.InputFiles
+import org.gradle.api.tasks.SkipWhenEmpty
 import org.jetbrains.kotlin.gradle.plugin.KotlinPlatformType
 import org.jetbrains.kotlin.gradle.plugin.KotlinTarget
 import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
@@ -51,5 +52,6 @@ public class KlibDumpMetadata(
      * Path to a resulting dump file.
      */
     @get:InputFiles
+    @get:SkipWhenEmpty
     public val dumpFile: RegularFileProperty
 ) : Serializable
--- a/src/main/kotlin/-Utils.kt
+++ b/src/main/kotlin/-Utils.kt
@@ -8,9 +8,7 @@ package kotlinx.validation
 import kotlinx.validation.api.klib.KlibTarget
 import kotlinx.validation.api.klib.konanTargetNameMapping
 import org.gradle.api.file.RegularFileProperty
-import org.gradle.api.tasks.Input
-import org.gradle.api.tasks.InputFiles
-import org.gradle.api.tasks.SkipWhenEmpty
+import org.gradle.api.tasks.*
 import org.jetbrains.kotlin.gradle.plugin.KotlinPlatformType
 import org.jetbrains.kotlin.gradle.plugin.KotlinTarget
 import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
@@ -50,8 +48,15 @@ public class KlibDumpMetadata(
 
     /**
      * Path to a resulting dump file.
+     *
+     * If a dump file was not generated for a particular [target] (which may
+     * happen for an empty project, a project having only test targets,
+     * or a project that has no sources for a particular target),
+     * [KlibDumpMetadata] will not force an error.
+     * Instead, a dependent task will be skipped.
      */
     @get:InputFiles
     @get:SkipWhenEmpty
+    @get:PathSensitive(PathSensitivity.RELATIVE)
     public val dumpFile: RegularFileProperty
 ) : Serializable
--- a/src/main/kotlin/-Utils.kt
+++ b/src/main/kotlin/-Utils.kt
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2016-2024 JetBrains s.r.o.
+ * Use of this source code is governed by the Apache 2.0 License that can be found in the LICENSE.txt file.
+ */
+
+package kotlinx.validation
+
+import kotlinx.validation.api.klib.KlibTarget
+import kotlinx.validation.api.klib.konanTargetNameMapping
+import org.gradle.api.file.RegularFileProperty
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.InputFiles
+import org.jetbrains.kotlin.gradle.plugin.KotlinPlatformType
+import org.jetbrains.kotlin.gradle.plugin.KotlinTarget
+import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
+import org.jetbrains.kotlin.gradle.targets.js.KotlinWasmTargetType
+import org.jetbrains.kotlin.gradle.targets.js.ir.KotlinJsIrTarget
+import java.io.Serializable
+
+/**
+ * Converts [KotlinTarget] to a [KlibTarget].
+ */
+public fun KotlinTarget.toKlibTarget(): KlibTarget = KlibTarget(extractUnderlyingTarget(this), targetName)
+
+private fun extractUnderlyingTarget(target: KotlinTarget): String {
+    if (target is KotlinNativeTarget) {
+        return konanTargetNameMapping[target.konanTarget.name]!!
+    }
+    return when (target.platformType) {
+        KotlinPlatformType.js -> "js"
+        KotlinPlatformType.wasm -> when ((target as KotlinJsIrTarget).wasmTargetType) {
+            KotlinWasmTargetType.WASI -> "wasmWasi"
+            KotlinWasmTargetType.JS -> "wasmJs"
+            else -> throw IllegalStateException("Unreachable")
+        }
+        else -> throw IllegalArgumentException("Unsupported platform type: ${target.platformType}")
+    }
+}
+
+/**
+ * Information about a generated klib dump.
+ */
+public class GeneratedDump(
+    /**
+     * The target the dump was generated for.
+     */
+    @get:Input
+    public val target: KlibTarget,
+
+    /**
+     * Path to a resulting dump file.
+     */
+    @get:InputFiles
+    public val dumpFile: RegularFileProperty
+) : Serializable
--- a/src/main/kotlin/-Utils.kt
+++ b/src/main/kotlin/-Utils.kt
@@ -40,7 +40,7 @@ private fun extractUnderlyingTarget(target: KotlinTarget): String {
 /**
  * Information about a generated klib dump.
  */
-public class GeneratedDump(
+public class KlibDumpMetadata(
     /**
      * The target the dump was generated for.
      */
--- a/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
+++ b/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
@@ -511,6 +511,23 @@ private class KlibValidationPipelineBuilder(
         val kotlin = project.kotlinMultiplatform
 
         val supportedTargetsProvider = supportedTargets()
+        val generatedDumps = objects.listProperty(GeneratedDump::class.java)
+        val inferredDumps = objects.listProperty(GeneratedDump::class.java)
+        mergeTask.configure {
+            it.dumps.addAll(generatedDumps)
+            it.doFirst {
+                if (supportedTargetsProvider.get().isEmpty()) {
+                    throw IllegalStateException(
+                        "KLib ABI dump/validation requires at least one enabled klib target, but none were found."
+                    )
+                }
+            }
+        }
+        mergeInferredTask.configure {
+            it.dumps.addAll(generatedDumps)
+            it.dumps.addAll(inferredDumps)
+        }
+
         kotlin.targets.matching { it.emitsKlib }.configureEach { currentTarget ->
             val mainCompilation = currentTarget.mainCompilationOrNull ?: return@configureEach
 
@@ -521,14 +538,9 @@ private class KlibValidationPipelineBuilder(
             // If a target is supported, the workflow is simple: create a dump, then merge it along with other dumps.
             if (targetSupported) {
                 val buildTargetAbi = configureKlibCompilation(mainCompilation, extension, targetConfig, apiBuildDir)
-                mergeTask.configure {
-                    it.dumps.add(GeneratedDump(targetName,
-                        objects.fileProperty().fileProvider(buildTargetAbi.map { it.outputApiFile })))
-                }
-                mergeInferredTask.configure {
-                    it.dumps.add(GeneratedDump(targetName,
-                        objects.fileProperty().fileProvider(buildTargetAbi.map { it.outputApiFile })))
-                }
+                generatedDumps.add(GeneratedDump(
+                    currentTarget.toKlibTarget(),
+                    objects.fileProperty().fileProvider(buildTargetAbi.map { it.outputApiFile })))
                 return@configureEach
             }
             // If the target is unsupported, the regular merge task will only depend on a task complaining about
@@ -542,22 +554,12 @@ private class KlibValidationPipelineBuilder(
             val proxy = unsupportedTargetDumpProxy(
                 mainCompilation,
                 klibApiDir, targetConfig,
-                KlibTarget(extractUnderlyingTarget(currentTarget), currentTarget.targetName),
-                apiBuildDir, supportedTargetsProvider
+                currentTarget.toKlibTarget(),
+                apiBuildDir
             )
-            mergeInferredTask.configure {
-                it.dumps.add(GeneratedDump(targetName,
-                    objects.fileProperty().fileProvider(proxy.map { it.outputFile })))
-            }
-        }
-        mergeTask.configure {
-            it.doFirst {
-                if (supportedTargetsProvider.get().isEmpty()) {
-                    throw IllegalStateException(
-                        "KLib ABI dump/validation requires at least one enabled klib target, but none were found."
-                    )
-                }
-            }
+            proxy.configure { it.dumps.addAll(generatedDumps) }
+            inferredDumps.add(GeneratedDump(currentTarget.toKlibTarget(),
+                objects.fileProperty().fileProvider(proxy.map { it.outputFile })))
         }
     }
 
@@ -584,7 +586,7 @@ private class KlibValidationPipelineBuilder(
                         true
                     }
                 }
-                .map { KlibTarget(extractUnderlyingTarget(it), it.targetName) }
+                .map { it.toKlibTarget() }
                 .toSet()
         }
     }
@@ -639,8 +641,7 @@ private class KlibValidationPipelineBuilder(
         klibApiDir: Provider<File>,
         targetConfig: TargetConfig,
         unsupportedTarget: KlibTarget,
-        apiBuildDir: File,
-        supportedTargets: Provider<Set<KlibTarget>>
+        apiBuildDir: File
     ): TaskProvider<KotlinKlibInferAbiForUnsupportedTargetTask> {
         val targetName = targetConfig.targetName!!
         return project.task<KotlinKlibInferAbiForUnsupportedTargetTask>(targetConfig.apiTaskName("Infer")) {
@@ -650,12 +651,9 @@ private class KlibValidationPipelineBuilder(
             description = "Try to infer the dump for unsupported target $targetName using dumps " +
                     "generated for supported targets."
             group = "other"
-            this.supportedTargets.addAll(supportedTargets)
             inputImageFile = klibApiDir.get().resolve(klibDumpFileName)
-            outputApiDir = apiBuildDir.toString()
             outputFile = apiBuildDir.resolve(klibDumpFileName)
             this.unsupportedTarget = unsupportedTarget
-            dumpFileName = klibDumpFileName
             dependsOn(project.tasks.withType(KotlinKlibAbiBuildTask::class.java))
         }
     }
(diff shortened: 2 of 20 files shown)


src/main/kotlin/KotlinKlibAbiBuildTask.kt
```diff
@@ -5,66 +5,62 @@
...
+public abstract class KotlinKlibAbiBuildTask : BuildTaskBase() {
 
     /**
-     * Path to a klib to dump.
+     * Collection consisting of a single path to a compiled klib (either file, or directory).
      */
-    @InputFiles
-    lateinit var klibFile: FileCollection
+    @get:InputFiles
+    @get:SkipWhenEmpty
```
shanshin, line 27: Is it expected situation when where is no file, `compilation.output.classesDirs` can be empty?

src/main/kotlin/KotlinKlibExtractAbiTask.kt
```diff
@@ -6,63 +6,65 @@
...
-    @get:Internal
-    internal val projectName = project.name
-
+public abstract class KotlinKlibExtractAbiTask : DefaultTask() {
     /**
      * Merged KLib dump that should be filtered by this task.
      */
-    @InputFiles
-    lateinit var inputAbiFile: File
+    @get:InputFiles
```
shanshin, line 27: - No `PathSensitive`

src/main/kotlin/KotlinKlibExtractAbiTask.kt
```diff
@@ -6,63 +6,65 @@
...
+    public abstract val inputAbiFile: RegularFileProperty
 
     /**
-     * A path to the resulting dump file.
+     * List of the targets that the resulting dump should contain.
      */
-    @OutputFile
-    lateinit var outputAbiFile: File
+    @get:Input
+    public abstract val requiredTargets: ListProperty<KlibTarget>
```
shanshin, line 34: If order is not import then it's worth to use  `SetProperty` (but in this case `KlibTarget` must have overridden equals/hashCode)

Optionally you may use `MapPropety<String, KlibTarget>` with `targetName -> KlibTarget`

src/main/kotlin/KotlinKlibInferAbiTask.kt
```diff
@@ -22,90 +23,70 @@ import java.io.File
...
-    @get:Internal
-    internal val projectName = project.name
-
-    /**
-     * The name of a target to infer a dump for.
-     */
-    @Input
-    lateinit var unsupportedTargetName: String
-
+public abstract class KotlinKlibInferAbiTask : DefaultTask() {
```
shanshin, line 26: Should be cacheable task

src/main/kotlin/KotlinKlibInferAbiTask.kt
```diff
@@ -22,90 +23,70 @@ import java.io.File
...
-    @InputFiles
-    lateinit var outputApiDir: String
-
-    /**
-     * Set of all supported targets.
-     */
-    @Input
-    lateinit var supportedTargets: Provider<Set<String>>
+    @get:Nested
+    public abstract val inputDumps: ListProperty<KlibDumpMetadata>
```
shanshin, line 37: Same about `SetProperty` or `MapProperty`

src/main/kotlin/KotlinKlibInferAbiTask.kt
```diff
@@ -22,90 +23,70 @@ import java.io.File
...
      */
-    @InputFiles
-    lateinit var inputImageFile: File
-
-    /**
-     * The name of a dump file.
-     */
-    @Input
-    lateinit var dumpFileName: String
+    @get:InputFiles
```
shanshin, line 42: - No `PathSensitive`

src/main/kotlin/KotlinKlibMergeAbiTask.kt
```diff
@@ -8,57 +8,39 @@ package kotlinx.validation
...
-    val targets: Set<String>
-        get() = targetToFile.keys
-
-    // Required to enforce task rerun on klibs update
-    @Suppress("UNUSED")
-    @get:InputFiles
-    internal val inputDumps: Collection<File>
-        get() = targetToFile.values
-
+public abstract class KotlinKlibMergeAbiTask : DefaultTask() {
```
shanshin, line 18: Should be cacheable

src/main/kotlin/KotlinKlibMergeAbiTask.kt
```diff
@@ -8,57 +8,39 @@ package kotlinx.validation
...
-     * A path to a resulting merged dump.
+     * Dumps to merge.
+     *
+     * If a file referred by [KlibDumpMetadata.dumpFile] does not exist, it will be ignored and corresponding
+     * target will not be mentioned in the resulting merged dump.
      */
-    @OutputFile
-    lateinit var mergedFile: File
+    @get:Nested
+    public abstract val dumps: ListProperty<KlibDumpMetadata>
```
shanshin, line 26: Same with `SetProperty` or `MapProperty`

src/main/kotlin/SyncFile.kt
```diff
@@ -6,26 +6,34 @@
...
 import java.nio.file.StandardCopyOption
 
 // Built-in Gradle's Copy/Sync tasks accepts only a destination directory (not a single file)
 // and registers it as an output dependency. If there's another task reading from that particular
 // directory or writing into it, their input/output dependencies would clash and as long as
 // there will be no explicit ordering or dependencies between these tasks, Gradle would be unhappy.
-internal open class CopyFile : DefaultTask() {
-    @InputFiles
-    lateinit var from: File
+internal abstract class SyncFile : DefaultTask() {
```
shanshin, line 21: Should be cacheable

src/main/kotlin/SyncFile.kt
```diff
@@ -6,26 +6,34 @@
...
+    @get:InputFiles
```
shanshin, line 22: - No `PathSensitive`

src/main/kotlin/SyncFile.kt
```diff
@@ -6,26 +6,34 @@
...
+    @get:OutputFile
+    abstract val to: RegularFileProperty
 
     @TaskAction
     fun copy() {
-        Files.copy(from.toPath(), to.toPath(), StandardCopyOption.REPLACE_EXISTING)
+        val fromFile = from.asFile.get()
+        val toFile = to.asFile.get()
+        if (fromFile.exists()) {
+            toFile.parentFile.mkdirs()
```
shanshin, line 33: What about `fromFile.copyTo(toFile, overwrite = true)`?
parent directories will be created automatically

src/main/kotlin/KotlinKlibInferAbiTask.kt
shanshin, line 42: File specified in `oldMergedKlibDump`  may not exist? Why we need `InputFiles`

src/main/kotlin/KotlinKlibExtractAbiTask.kt
shanshin, line 27: File specified in `inputAbiFile`  may not exist? Why we need `InputFiles`

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -388,76 +385,56 @@ private class KlibValidationPipelineBuilder(
...
         group = "other"
-        strictValidation = extension.klib.strictValidation
-        supportedTargets = supportedTargets()
-        inputAbiFile = klibApiDir.get().resolve(klibDumpFileName)
-        outputAbiFile = klibOutputDir.resolve(klibDumpFileName)
-        val hasCompilableTargets = project.hasCompilableTargetsPredicate()
-        onlyIf("There are no klibs compiled for the project") { hasCompilableTargets.get() }
+        strictValidation.set(extension.klib.strictValidation)
+        requiredTargets.addAll(supportedTargets())
+        inputAbiFile.set(klibApiDir.get().resolve(klibDumpFileName))
```
shanshin, line 436: `klibApiDir.map { it.resolve(...) }` ?

src/main/kotlin/-Utils.kt
```diff
@@ -0,0 +1,57 @@
...
+     */
+    @get:Input
+    public val target: KlibTarget,
+
+    /**
+     * Path to a resulting dump file.
+     */
+    @get:InputFiles
+    @get:SkipWhenEmpty
+    public val dumpFile: RegularFileProperty
```
shanshin, line 56: `Provider<File>`?
Properties is not very often used in nested types, in this case, you will not need to call `objects.fileProperty()`

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -570,54 +538,42 @@ private class KlibValidationPipelineBuilder(
...
-            onlyIf { hasSourcesPredicate.get() }
             // 'group' is not specified deliberately, so it will be hidden from ./gradlew tasks
             description = "Builds Kotlin KLib ABI dump for 'main' compilations of $projectName. " +
                     "Complementary task and shouldn't be called manually"
-            klibFile = project.files(project.provider { compilation.output.classesDirs })
-            compilationDependencies = project.files(project.provider { compilation.compileDependencyFiles })
-            signatureVersion = SerializableSignatureVersion(extension.klib.signatureVersion)
-            outputApiFile = apiBuildDir.resolve(klibDumpFileName)
+            this.target.set(target)
+            klibFile.from(project.provider { compilation.output.classesDirs })
```
shanshin, line 574: `ConfigurableFileCollection` is a live collection, try use it directly `klibFile.from(compilation.output.classesDirs )`

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -636,29 +592,21 @@ private class KlibValidationPipelineBuilder(
     }
 
     private fun Project.unsupportedTargetDumpProxy(
-        compilation: KotlinCompilation<KotlinCommonOptions>,
+        @Suppress("UNUSED_PARAMETER") compilation: KotlinCompilation<KotlinCommonOptions>,
```
shanshin, line 595: Why unused parameter is still present?

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -636,29 +592,21 @@ private class KlibValidationPipelineBuilder(
...
-            this.supportedTargets = supportedTargets
-            inputImageFile = klibApiDir.get().resolve(klibDumpFileName)
-            outputApiDir = apiBuildDir.toString()
-            outputFile = apiBuildDir.resolve(klibDumpFileName)
-            unsupportedTargetName = targetConfig.targetName
-            unsupportedTargetCanonicalName = underlyingTarget
-            dumpFileName = klibDumpFileName
-            dependsOn(project.tasks.withType(KotlinKlibAbiBuildTask::class.java))
+            target.set(unsupportedTarget)
+            oldMergedKlibDump.set(klibApiDir.get().resolve(klibDumpFileName))
```
shanshin, line 608: `oldMergedKlibDump.fileProvider(klibApiDir.map { it.resolve(...) }`

src/main/kotlin/KotlinApiCompareTask.kt
```diff
@@ -63,17 +67,17 @@ public open class KotlinApiCompareTask @Inject constructor(private val objects:
             expectedApiFiles[file.name] = file.relativePath
```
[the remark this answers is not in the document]
fzhinkin: Let me revisit the original change that was aimed to overcome the issue with case-insensitive filesystems

src/main/kotlin/KotlinKlibExtractAbiTask.kt
fzhinkin: Thanks for pointing. Yeap, the order does not matter here, and, thankfully, `KlibTarget` already overridden equals and hashCode.

src/main/kotlin/-Utils.kt
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * The target the dump was generated for.
+     */
+    @get:Input
+    public val target: KlibTarget,
+
+    /**
+     * Path to a resulting dump file.
+     */
+    @get:InputFiles
```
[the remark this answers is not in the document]
fzhinkin: Update KDoc with the details on why a file could be missing and what to expect then

src/main/kotlin/-Utils.kt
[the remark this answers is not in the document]
fzhinkin: Fixed, thanks!

src/main/kotlin/KotlinApiCompareTask.kt
```diff
@@ -17,25 +17,29 @@ import org.gradle.api.tasks.*
 
 public open class KotlinApiCompareTask @Inject constructor(private val objects: ObjectFactory): DefaultTask() {
```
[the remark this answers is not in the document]
fzhinkin: Fixed, thanks!

src/main/kotlin/KotlinKlibAbiBuildTask.kt
fzhinkin: Yes, it is. Updated KDoc

src/main/kotlin/KotlinApiBuildTask.kt
```diff
@@ -9,11 +9,14 @@ import kotlinx.validation.api.*
...
 import org.gradle.api.file.*
 import org.gradle.api.tasks.*
+import java.io.File
 import java.util.jar.JarFile
 import javax.inject.Inject
 
 public open class KotlinApiBuildTask @Inject constructor(
 ) : BuildTaskBase() {
+    @OutputFile
+    public lateinit var outputApiFile: File
```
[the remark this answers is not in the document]
shanshin: You can try to add a reliable field, but maintain backward compatibility.
In `BuildTaskBase`
```
    @get:Internal
    public abstract var outputApiFile: File
```

In `KotlinApiBuildTask`
```
    @get:OutputFile
    ...
    public val realOutputApiFile: RegularFileProperty = project.objects.fileProperty()

    @get:Internal
    public override var outputApiFile: File
        get() = realOutputApiFile.get().asFile
        set(value) = realOutputApiFile.set(value)
```

* I didn 't check it )

src/main/kotlin/KotlinKlibAbiBuildTask.kt
```diff
@@ -5,66 +5,62 @@
...
     internal fun generate() {
-        outputApiFile.delete()
-        outputApiFile.parentFile.mkdirs()
+        val outputFile = outputAbiFile.asFile.get()
+        outputFile.delete()
+        outputFile.parentFile.mkdirs()
 
-        val dump = KlibDump.fromKlib(klibFile.singleFile, target, KLibDumpFilters {
+        val dump = KlibDump.fromKlib(klibFile.singleFile, target.get().configurableName, KLibDumpFilters {
             ignoredClasses.addAll(this@KotlinKlibAbiBuildTask.ignoredClasses)
```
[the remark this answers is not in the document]
shanshin: See comment about potential compatibility

src/main/kotlin/KotlinApiBuildTask.kt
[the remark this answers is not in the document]
fzhinkin: The main issue here is that we definitely want to phase these old properties out and ask users to migrate to `realXXX` ones, but old properties already claimed good names :(

src/main/kotlin/KotlinApiCompareTask.kt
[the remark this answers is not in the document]
fzhinkin: So, I took a look at the problem once again, and it seems like the best option would be to remove all case-insensitive file name logic from BCV (https://github.com/Kotlin/binary-compatibility-validator/issues/231).

Otherwise, all tasks have to use directories as their inputs/outputs and a filename specified separately as an additional input. Moreover, all tasks processing API files from the project repo (a.k.a. golded files) have to also implement case-insensitive file lookup, the same way compare task do that.

src/main/kotlin/-Utils.kt
```diff
@@ -0,0 +1,57 @@
...
+    @get:SkipWhenEmpty
```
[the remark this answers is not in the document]
fzhinkin: Yes, KDoc updated with the details

src/main/kotlin/KotlinApiCompareTask.kt
```diff
@@ -17,25 +17,29 @@ import org.gradle.api.tasks.*
...
 
-    @InputFiles
-    @PathSensitive(PathSensitivity.RELATIVE)
-    public lateinit var projectApiFile: File
+    @get:InputFiles
+    @get:SkipWhenEmpty
```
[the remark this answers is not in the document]
fzhinkin: It could be empty so we can emit a more user-friendly error message.
However, adding `SkipWhenEmpty` was a mistake: KLib validation should behave the same way JVM-ABI validation does, and the latter fails without a `.api` file, even for empty projects.

src/main/kotlin/KotlinApiCompareTask.kt
```diff
@@ -17,25 +17,29 @@ import org.gradle.api.tasks.*
...
+    @get:PathSensitive(PathSensitivity.RELATIVE)
+    public val projectApiFile: RegularFileProperty = objects.fileProperty()
 
-    @InputFiles
-    public lateinit var generatedApiFile: File
+    @get:InputFiles
```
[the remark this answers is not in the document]
fzhinkin: Yes, for error reporting purposes only

src/main/kotlin/KotlinApiBuildTask.kt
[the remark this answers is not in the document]
shanshin: Will changing property type break a lot? Because if we don't change it now, it will stay that way for a long time.

src/main/kotlin/KotlinApiCompareTask.kt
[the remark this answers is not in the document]
shanshin: My bad)
This task has no output, so it shouldn't be cacheable

src/main/kotlin/KotlinApiCompareTask.kt
```diff
@@ -10,68 +10,69 @@ import com.github.difflib.UnifiedDiffUtils
 import java.io.*
 import javax.inject.Inject
 import org.gradle.api.*
+import org.gradle.api.file.RegularFileProperty
 import org.gradle.api.tasks.*
 
-public open class KotlinApiCompareTask @Inject constructor(): DefaultTask() {
+@CacheableTask
+public open class KotlinApiCompareTask @Inject constructor() : DefaultTask() {
```
shanshin, line 17: `@Inject constructor()` is not needed in this case

src/main/kotlin/SyncFile.kt
shanshin: I thought again about this task, since it does not perform any complex calculations, then you should not inflate the 
 build cache.

It is better to remove this annotation and put `@DisableCachingByDefault(because = "No computations, only copying files")`

src/main/kotlin/KotlinApiBuildTask.kt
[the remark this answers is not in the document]
fzhinkin: A few weeks ago I had an impression that there are a lot of projects relying on these properties. After checking it once again, it seems like I was wrong and there are only a couple of projects on GH relying on it.

Originally, I was reluctant to make breaking changes as I wanted to limit the scope of the release to KLib support only. But as there are some other breaking changes already, it might be worth changing this part too and provide a migration guide for those who use corresponding properties.

As it was previously discussed elsewhere, currently BCV is mostly concerned with its own source compatibility as there are no other projects (beside Adam's BCV-MU) extending the BCV, so it should be more or less fine to change property types now.

src/main/kotlin/BuildTaskBase.kt
```diff
@@ -6,52 +6,43 @@
...
-        get() = _nonPublicMarkes ?: extension?.nonPublicMarkers ?: emptySet()
-        set(value) { _nonPublicMarkes = value }
+    public val nonPublicMarkers: SetProperty<String> = stringSetProperty { nonPublicMarkers }
 
-    private var _ignoredClasses: Set<String>? = null
     @get:Input
-    public var ignoredClasses : Set<String>
-        get() = _ignoredClasses ?: extension?.ignoredClasses ?: emptySet()
-        set(value) { _ignoredClasses = value }
+    public var ignoredClasses: SetProperty<String> = stringSetProperty { ignoredClasses }
```
shanshin, line 36: Should be `val`

src/main/kotlin/BuildTaskBase.kt
```diff
@@ -6,52 +6,43 @@
...
 
-    private var _publicPackages: Set<String>? = null
     @get:Input
-    public var publicPackages: Set<String>
-        get() = _publicPackages ?: extension?.publicPackages ?: emptySet()
-        set(value) { _publicPackages = value }
+    public var publicPackages: SetProperty<String> = stringSetProperty { publicPackages }
```
shanshin, line 39: Should be `val`

src/main/kotlin/BuildTaskBase.kt
```diff
@@ -6,52 +6,43 @@
...
 
-    private var _publicMarkers: Set<String>? = null
     @get:Input
-    public var publicMarkers: Set<String>
-        get() = _publicMarkers ?: extension?.publicMarkers ?: emptySet()
-        set(value) { _publicMarkers = value}
+    public var publicMarkers: SetProperty<String> = stringSetProperty { publicMarkers }
```
shanshin, line 42: Should be `val`

src/main/kotlin/BuildTaskBase.kt
```diff
@@ -6,52 +6,43 @@
...
 
-    private var _publicClasses: Set<String>? = null
     @get:Input
-    public var publicClasses: Set<String>
-        get() = _publicClasses ?: extension?.publicClasses ?: emptySet()
-        set(value) { _publicClasses = value }
+    public var publicClasses: SetProperty<String> = stringSetProperty { publicClasses }
```
shanshin, line 45: Should be `val`

src/main/kotlin/BuildTaskBase.kt
```diff
@@ -6,52 +6,43 @@
...
 
 import org.gradle.api.DefaultTask
+import org.gradle.api.provider.SetProperty
 import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.Internal
-import org.gradle.api.tasks.OutputFile
-import java.io.File
+
 
 public abstract class BuildTaskBase : DefaultTask() {
```
shanshin, line 14: Should we link a public task so tightly to a specific extension?

shanshin (approved):

src/main/kotlin/BuildTaskBase.kt
fzhinkin: These tasks are tightly bound to BCV anyway. The thing that worries me more is implicit initialization of properties, but I'm not sure that making initialization explicit and moving it to `BinaryCompatibilityValidatorPlugin` will make a big difference for now

PR #212, opened 2024-04-02
Revert JVM tasks configuration changes introduced in #200

fzhinkin:


fzhinkin:
The proper solution require linking tasks with each other using input and output files, that will be done separately. For now, I'll just revert recent changes breaking the behavior.

shanshin (approved):

PR #226, opened 2024-05-09
allow apiDumpDirectory to be outside project dir as long as inside rootProject dir

robstoll:
Shall be understood as pre-work, I guess that there are test which will fail now (as the check changed) and additional tests would be good which still ensure that we cannot write outside the rootProject dir.

This change allows that e.g. in an multimodule project all api files are written to a single place within the rootProject dir (in my case I would like to put everything into rootProjectDir/misc/api and use `apiDumpDirectory = rootProject.projectDir.relativeTo(project.projectDir).resolve("misc/api").toString()`


robstoll:
missed that `develop` is the branch we should put PRs against. going to open another PR

PR #227, opened 2024-05-09
allow apiDumpDirectory outside projectDir if inside rootProjectDir

robstoll:
Shall be understood as pre-work, I guess that there are test which will fail now (as the check changed) and additional tests would be good which still ensure that we cannot write outside the rootProject dir.

This change allows that e.g. in an multimodule project all api files are written to a single place within the rootProject dir (in my case I would like to put everything into rootProjectDir/misc/api and use `apiDumpDirectory = rootProject.projectDir.relativeTo(project.projectDir).resolve("misc/api").toString()`

### DIFF
--- a/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
+++ b/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
@@ -162,13 +162,14 @@ private class TargetConfig constructor(
     private val apiDirProvider = project.provider {
         val dir = extension.apiDumpDirectory
 
-        val root = project.layout.projectDirectory.asFile.toPath().toAbsolutePath().normalize()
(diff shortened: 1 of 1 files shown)


fzhinkin:
Hey @robstoll! Could you please elaborate on your plans regarding this PR?

I waited for a while because of:
> Shall be understood as pre-work

But it feels like the time to ask has come :)

robstoll:
Sorry, wasn't clear enough. Shall be understood as pre-work for a maintainer to take over.

fzhinkin:
@robstoll, if you're willing to add a new feature to the BCV, please follow [contributing guidelines](https://github.com/Kotlin/binary-compatibility-validator/blob/master/CONTRIBUTING.md) and open an issue describing why it's needed, what problems it solves, and how it is supposed to work. If there will be a consensus that the feature should be supported, maybe someone will take it over and implement it.

PR #236, opened 2024-06-05
Correctly handle disappeared native/web targets

### LINKED ISSUE #234: Removing native targets did not cause API check to fail
I removed a bunch of multiplatform targets in https://github.com/cashapp/redwood/pull/2070 but made no actual API changes.

There were three types of removals:

1. Module with ios (arm/x86/simulator) and macos (arm/x86) were changed to be ios-only.
2. Module with ios, macos, jvm, and js was changed to jvm and js only.
3. Module with android and jvm were changed to be jvm only.

In all three cases the `apiCheck` task did not fail despite running `apiDump` producing _significant_ changes to the API files (as seen in https://github.com/cashapp/redwood/pull/2077).

Points 1 and 2 are similar, both being cases of the targets in the API file being now incorrect. Point 3 is interesting because the API files in `android/` and `jvm/` were now dead and the API file in the root was missing.

I would have hoped all three cases to be detected, fail the build, and require an `apiDump` to update the files.

fzhinkin:
To handle cases when a dump containing declarations for Apple targets is validated on hosts not supporting these targets, BCV filters out a golden dump committed to the repo. 

Unfortunately, the filtration logic was invalid. Instead of filtering out targets known to be unsupported, the corresponding task was filtering out everything except supported targets. If a target was removed from the config, it could no longer make it to a set of supported/enabled targets, and the aforementioned task would filter it out.

Fix changes the filtration logic only to remove targets that were defined in a project but are known to be unsupported.

Fixes #234

### DIFF
--- a/src/functionalTest/kotlin/kotlinx/validation/test/KlibVerificationTests.kt
+++ b/src/functionalTest/kotlin/kotlinx/validation/test/KlibVerificationTests.kt
@@ -791,4 +791,26 @@ internal class KlibVerificationTests : BaseKotlinGradleTest() {
             assertTaskFailure(":klibApiCheck")
         }
     }
+
+    @Test
+    fun `apiCheck should fail after target removal`() {
+        val runner = test {
+            settingsGradleKts {
+                resolve("/examples/gradle/settings/settings-name-testproject.gradle.kts")
+            }
+            // only a single native target is defined there
+            buildGradleKts {
+                resolve("/examples/gradle/base/withNativePluginAndSingleTarget.gradle.kts")
+            }
+            addToSrcSet("/examples/classes/AnotherBuildConfig.kt")
+            // dump was created for multiple native targets
+            abiFile(projectName = "testproject") {
+                resolve("/examples/classes/AnotherBuildConfig.klib.dump")
+            }
+            runApiCheck()
+        }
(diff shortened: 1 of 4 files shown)


src/functionalTest/kotlin/kotlinx/validation/test/KlibVerificationTests.kt
```diff
@@ -752,4 +752,26 @@ internal class KlibVerificationTests : BaseKotlinGradleTest() {
...
+            }
+            addToSrcSet("/examples/classes/AnotherBuildConfig.kt")
+            // dump was created for multiple native targets
+            abiFile(projectName = "testproject") {
+                resolve("/examples/classes/AnotherBuildConfig.klib.dump")
+            }
+            runApiCheck()
+        }
+        runner.buildAndFail().apply {
+            assertTaskFailure(":klibApiCheck")
```
shanshin, line 774: Shouldn't we check the message in some way?

shanshin (approved):

src/functionalTest/kotlin/kotlinx/validation/test/KlibVerificationTests.kt
fzhinkin: Yes, thanks! I improved the check.

PR #237, opened 2024-06-05
Stop supporting case insensitive file names

### LINKED ISSUE #231: Remove case-insensitive dump file names handling logic
https://github.com/Kotlin/binary-compatibility-validator/pull/79 and https://github.com/Kotlin/binary-compatibility-validator/pull/84 addressed the issue reported in https://github.com/Kotlin/binary-compatibility-validator/issues/76 by looking up `.api` files in a case-insensitive manner.

To make things work smoothly on the Gradle side, we need to:
- keep track of individual files instead of a directory containing these files (which is not possible with how we supported case-insensitivity)
- keep track of directories (for check tasks, it's how we did it previously).

The latter option should work, but it requires specifying input/output directories and a file name for all the tasks. Otherwise, Gradle won't be able to track dependencies between tasks correctly. Also, all tasks consuming `.api` files from the repo should be updated to support case-insensitive file names (and for Klibs, the majority of tasks need it).

Note that before all the klib-related changes, the check task was using directories as its inputs, while the dump task was using files as its output, so the only way to enforce dependency between these tasks was `dependsOn` call.

Looking at #76 I'm not sure if macOS's case insensitivity was the root cause. The `:apiDump` task was based on `Sync`, which wipes out the destination directory before copying files. Even if the dump had a name with the wrong characters case, rerunning `apiDump` should have fixed it.

So, I propose removing all the logic for case-insensitive file names from the BCV and using exact file paths as Gradle task inputs/outputs.

fzhinkin:
This PR fixes only behaviour, task's API will be cleaned up in #204.

Fixes #231

### DIFF
--- a/src/functionalTest/kotlin/kotlinx/validation/test/DefaultConfigTests.kt
+++ b/src/functionalTest/kotlin/kotlinx/validation/test/DefaultConfigTests.kt
@@ -7,7 +7,10 @@ package kotlinx.validation.test
 
 import kotlinx.validation.api.*
 import org.assertj.core.api.*
+import org.junit.Assume
 import org.junit.Test
+import java.io.File
+import java.nio.file.Files
 import kotlin.test.*
 
 internal class DefaultConfigTests : BaseKotlinGradleTest() {
@@ -90,7 +93,9 @@ internal class DefaultConfigTests : BaseKotlinGradleTest() {
     }
 
     @Test
-    fun `apiCheck should succeed when public classes match api file ignoring case`() {
+    fun `apiCheck should fail when public classes match api file ignoring case`() {
+        Assume.assumeTrue(underlyingFsIsCaseSensitive())
+
         val runner = test {
             buildGradleKts {
                 resolve("/examples/gradle/base/withPlugin.gradle.kts")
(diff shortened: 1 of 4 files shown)


src/main/kotlin/KotlinApiCompareTask.kt
```diff
@@ -41,43 +38,17 @@ public open class KotlinApiCompareTask @Inject constructor(private val objects:
         }
         val subject = projectName
 
-        /*
```
shanshin, line 44: Hurray! )

src/functionalTest/kotlin/kotlinx/validation/test/DefaultConfigTests.kt
```diff
@@ -90,7 +93,9 @@ internal class DefaultConfigTests : BaseKotlinGradleTest() {
     }
 
     @Test
-    fun `apiCheck should succeed when public classes match api file ignoring case`() {
+    fun `apiCheck should fail when public classes match api file ignoring case`() {
+        Assume.assumeTrue(underlyingFsIsCaseSensitive())
```
shanshin, line 97: Are the tests run on different systems at once?

shanshin (approved):

src/functionalTest/kotlin/kotlinx/validation/test/DefaultConfigTests.kt
fzhinkin: They are running on Windows/Linux/MacOS from time to time, yes.

PR #238, opened 2024-06-06
Use foojay-resolver-convention to resolve JVM toolchains

fzhinkin:
Recently, I was running BCV tests on hosts with only recent Java installed, and automatic toolchain provisioning could made my life much easier.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -4,7 +4,6 @@ import kotlinx.validation.build.mavenRepositoryPublishing
 import kotlinx.validation.build.signPublicationIfKeyPresent
 import org.gradle.api.attributes.TestSuiteType.FUNCTIONAL_TEST
 import org.jetbrains.dokka.gradle.DokkaTask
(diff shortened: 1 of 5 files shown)


fzhinkin:
> Why use a toolchain when any modern JDK can cross-compile to Java 8 with faster compilation speeds and bytecode generation bugfixes using the --release flag?

@JakeWharton, it's just more convenient compared to explicitly specifying the target for all kotlin- and java-tasks.

shanshin:
@JakeWharton, I would also prefer to use flags, however, not all of our agents have JDK installed on which the flag `-Xjdk-release` works. 

@fzhinkin, could you clarify a little bit why we are limiting the JDK version? To make the BCV plugin work if Gradle runs on older JVMs?

fzhinkin:
@shanshin, yes, the version is limited for Gradle compatibility reasons (it works with Java >= 8).

shanshin (approved):

PR #244, opened 2024-06-25
Require JVM ABI dump presence in empty KMP projects

### LINKED ISSUE #243: BCV behavior for an empty module is inconsistent between JVM and KMP
When applied to K/JVM module, BCV require to create an API dump even if a module does not have any "production" sources (for instance, it's a test-only module).

At the same time, if it's KMP module, BCV simply ignores the module (instead of reporting an error and asking to create an empty dump).

KLib support complicated the matter:
- in 0.15.0-Beta.{1,2} such a module will be ignored, the same way it's done for JVM ABI validation;
- develop branch contains what should become 0.15.0-Beta.3 and KLib validation fails for an empty module (the same way it would fail for a K/JVM empty module).

BCV should validation should behave consistently for both JVM and KLib ABI as well as both K/JVM and KMP projects.
Current JVM ABI validation behavior for KMP projects seems to be incorrect as sources removal will lead to the `:apiCheck` task being skipped instead of it being executed and failed.

fzhinkin:
Currently, only JVM ABI validation in KMP projects allows having no dump file. In other cases (JVM ABI validation in K/JVM project or KLib validation) require the presence of an ABI dump file. Lack of such a file is treated as a validation error.

The main motivation for having a dump files instead of simply ignoring such projects is that otherwise BCV won't be able to catch the case when all sources were removed from a project.

Fixes #243

### DIFF
--- a/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
+++ b/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
@@ -214,7 +214,6 @@ private fun Project.configureKotlinCompilation(
     val apiBuildDir = apiDirProvider.flatMap { f -> layout.buildDirectory.asFile.map { it.resolve(f) } }
 
     val apiBuild = task<KotlinApiBuildTask>(targetConfig.apiTaskName("Build")) {
-        // Do not enable task for empty umbrella modules
         isEnabled = apiCheckEnabled(projectName, extension)
         // 'group' is not specified deliberately, so it will be hidden from ./gradlew tasks
         description =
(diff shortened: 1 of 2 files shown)


src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -215,7 +215,7 @@ private fun Project.configureKotlinCompilation(
 
     val apiBuild = task<KotlinApiBuildTask>(targetConfig.apiTaskName("Build")) {
         // Do not enable task for empty umbrella modules
```
shanshin, line 217: Is this comment still actual?

shanshin (approved):

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
fzhinkin: No, thanks for pointing! Removed this one and yet another, left in klib-related code.

Issue #90, opened 2022-06-22
Companion object 'const val' properties generate false positive if they have a non-public marker


martinbonnin:
The following:

```kotlin
@Target(AnnotationTarget.PROPERTY)
annotation class HiddenProperty

public class Foo {
    companion object {
        @HiddenProperty
        const val bar = "barValue"
    }
}
```

Generates the following pseudo Java code:

```java
public final class Foo {
   @NotNull
   public static final String bar = "barValue";
   @NotNull
   public static final Companion Companion = new Companion((DefaultConstructorMarker)null);

   public static final class Companion {
      // $FF: synthetic method
      @HiddenProperty
      public static void getBar$annotations() {
      }

      private Companion() {
      }
   }
}
```

Because the synthetic method carrying the annotations is not in the same class as the field, it is ignored and the field is marked as public API even though it shouldn't.

qwwdfsad:
I wonder how this translates to Java interop stability. But I guess if it's annotated and the documentation is straightforward, it shouldn't surprise anyone

martinbonnin:
Indeed, the Java story, without `@OptIn` annotations sounds more dangerous. I guess it's all going through documentation...

martinbonnin:
Just bumped into that again.

> I wonder how this translates to Java interop stability.

This is true for all non-public markers, right? Java code doesn't really "see" them so it's all based on documentation for java callers. Or maybe recommend library authors to add `@JvmSynthetic` to all non-public marker annotated symbols?

martinbonnin:
Attempt at fixing this: https://github.com/Kotlin/binary-compatibility-validator/pull/245

martinbonnin:
Closing as #245 was merged

PR #245, opened 2024-06-27
Fix false positive for const val with non-public marker

martinbonnin:
Attempt at fixing #90.

### DIFF
--- a/src/functionalTest/kotlin/kotlinx/validation/test/NonPublicMarkersTest.kt
+++ b/src/functionalTest/kotlin/kotlinx/validation/test/NonPublicMarkersTest.kt
@@ -104,4 +104,30 @@ class NonPublicMarkersTest : BaseKotlinGradleTest() {
             Assertions.assertThat(rootProjectApiDump.readText()).isEqualToIgnoringNewLines(expected)
         }
     }
+
+    @Test
+    fun testIgnoredMarkersOnConstProperties() {
+        val runner = test {
+            buildGradleKts {
+                resolve("/examples/gradle/base/withPlugin.gradle.kts")
+                resolve("/examples/gradle/configuration/nonPublicMarkers/markers.gradle.kts")
+            }
+
+            kotlin("ConstProperty.kt") {
+                resolve("/examples/classes/ConstProperty.kt")
+            }
+
+            apiFile(projectName = rootProjectDir.name) {
+                resolve("/examples/classes/ConstProperty.dump")
+            }
+
+            runner {
+                arguments.add(":apiCheck")
+            }
+        }
+
+        runner.withDebug(true).build().apply {
+            assertTaskSuccess(":apiCheck")
+        }
+    }
 }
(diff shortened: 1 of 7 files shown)


src/functionalTest/kotlin/kotlinx/validation/test/NonPublicMarkersTest.kt
```diff
@@ -104,4 +104,30 @@ class NonPublicMarkersTest : BaseKotlinGradleTest() {
             Assertions.assertThat(rootProjectApiDump.readText()).isEqualToIgnoringNewLines(expected)
         }
     }
+
+    @Test
+    fun testIgnoredMarkersOnConstProperties() {
```
fzhinkin, line 109: Could you please also extend the companions test here https://github.com/Kotlin/binary-compatibility-validator/tree/master/src/test/kotlin/cases/companions ?
It would be nice to cover cases with multiple annotated `const val` properties as well as the case where some of them are annotated and filtered, while others are not annotated and should be dumped.

src/main/kotlin/api/KotlinSignaturesLoading.kt
```diff
@@ -119,10 +122,26 @@ private fun FieldNode.buildFieldSignature(
          * as non-public API using some annotation, then we won't be able to filter out
          * the companion field.
          */
-        val companionName = ownerClass.companionName(ownerClass.kotlinMetadata)
+        val companionName = ownerClass.companionName(ownerClass.kotlinMetadata)!!
         companionClass = classes[companionName]
         foundAnnotations.addAll(companionClass?.visibleAnnotations.orEmpty())
         foundAnnotations.addAll(companionClass?.invisibleAnnotations.orEmpty())
+    } else if (access.and(Opcodes.ACC_STATIC) != 0 && access.and(Opcodes.ACC_FINAL) != 0) {
```
fzhinkin, line 129: `isStatic(access) && isFinal(access)`

fzhinkin (changes requested):
@martinbonnin, thanks for addressing the issue! Overall, the change looks good, I have just a few minor nits.

src/functionalTest/kotlin/kotlinx/validation/test/NonPublicMarkersTest.kt
martinbonnin: Done in https://github.com/Kotlin/binary-compatibility-validator/pull/245/commits/7764ffb34169afa9e958f3053c617f12899ddbbb, let me know what you think.

martinbonnin: I forgot to cover "multiple annotated `const val` properties". Added in https://github.com/Kotlin/binary-compatibility-validator/pull/245/commits/ab771fd73f0a8797395158dc569c028aa55a0814

fzhinkin (approved):
LGTM, thanks! I'll merge it later today

Issue #246, opened 2024-07-01
JVM/Android-only modules create empty klib.api file on 0.15.0-Beta.3


colinrtwhite:
On `0.15.0-Beta.3`, the plugin generates an empty `klib.api` if the [module only supports Android/JVM targets](https://github.com/coil-kt/coil/blob/main/coil-network-okhttp/build.gradle.kts#L16). This causes the binary compatibility check to fail with the error `Project ABI file XYZ.klib.api is empty.`. The release notes mention adding modules to `ignoredProjects`, but I believe that also disables checking the `android/XYZ.api` and `jvm/XYZ.api` files. I have the new klib validation support enabled with `strictValidation = true`:

```kotlin
klib {
    enabled = true
    strictValidation = true
}
```

To reproduce:

Run `./gradlew apiDump` and then `./gradlew apiCheck` in [this project](https://github.com/coil-kt/coil/pull/2343). I can make a smaller repro project if needed.

fzhinkin:
@colinrtwhite thanks for reporting!

PR #247, opened 2024-07-02
Align klib validation behavior for empty projects with other targets

fzhinkin:
After #243 resolution, JVM/JVM on KMP/KMP projects validation behavior remained different for projects without non-test sources or without published API:
- For JVM projects (of all kinds), we generate empty dump files and verify that the dump remained empty on `apiCheck` execution;
- For KMP KLib-targets, dump generation is skipped for such projects; the validation fails due to a missing dump file.

This PR addresses #246 and aligns KMP validation behavior for empty projects with how JVM ABI validation works for them.

### DIFF
--- a/src/main/kotlin/-Utils.kt
+++ b/src/main/kotlin/-Utils.kt
@@ -56,7 +56,6 @@ public class KlibDumpMetadata(
      * Instead, a dependent task will be skipped.
      */
     @get:InputFiles
-    @get:SkipWhenEmpty
     @get:PathSensitive(PathSensitivity.RELATIVE)
     public val dumpFile: RegularFileProperty
 ) : Serializable
--- a/src/main/kotlin/-Utils.kt
+++ b/src/main/kotlin/-Utils.kt
@@ -53,7 +53,8 @@ public class KlibDumpMetadata(
      * happen for an empty project, a project having only test targets,
      * or a project that has no sources for a particular target),
      * [KlibDumpMetadata] will not force an error.
-     * Instead, a dependent task will be skipped.
+     * It's up to [KlibDumpMetadata] users to check if a [dumpFile] exists and
+     * perform a required action accordingly.
      */
     @get:InputFiles
     @get:PathSensitive(PathSensitivity.RELATIVE)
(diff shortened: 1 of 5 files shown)


src/main/kotlin/-Utils.kt
```diff
@@ -56,7 +56,6 @@ public class KlibDumpMetadata(
      * Instead, a dependent task will be skipped.
```
shanshin, line 56: Comment isn't actual

src/main/kotlin/-Utils.kt
```diff
@@ -56,7 +56,6 @@ public class KlibDumpMetadata(
...
      */
     @get:InputFiles
-    @get:SkipWhenEmpty
     @get:PathSensitive(PathSensitivity.RELATIVE)
     public val dumpFile: RegularFileProperty
```
shanshin, line 60: If we removed `SkipWhenEmpty` and saved `InputFiles` so we assume that `dumpFile.isPresent` can be `false`. 
Is it checked somewhere below?

src/main/kotlin/-Utils.kt
fzhinkin: Thanks! Updated

src/main/kotlin/-Utils.kt
shanshin: No, that's not what I meant.

In fact, there are two levels of presence here: 
- `RegularFileProperty` is not absent (`dumpFile.set` or `dumpFile.convention` has been called anywhere at least once)
- the file from `dumpFile.get()` exists

Annotation `InputFiles` is specified, it is used when one, several, or zero files (case with `.isPresent == false`) can be added in a property.
If one and only one file is always written to the property, then it is worth replacing it with `InputFile`

fzhinkin: The reason behind `InputFiles` being used instead of `InputFile` is that there might be no file and `InputFile` results in an assertion that checks file's existence.

fzhinkin: If the `dumpFile` was not set, it's an error, so IAE raised on the `.get` call should be fine

shanshin (approved):

PR #251, opened 2024-07-08
Postpone static field's companion resolution

### LINKED ISSUE #250: [0.15.0] Enum entries are not part of generated `.api` dump anymore if the enum contains internal companion object
For a given class:
```kotlin
package com.luciad.imageio.webp

public enum class CompressionType(internal val imageIoValue: String) {
    Lossy("Lossy"),
    Lossless("Lossless"),
    ;

    internal companion object {

        val imageIoCompressionTypes = entries.map { it.imageIoValue }.toTypedArray()
    }
}
```

Migration from 0.14.0 to 0.15.0 produces the following build failure:
```java
Execution failed for task ':webp-imageio:apiCheck'.
> API check failed for project webp-imageio.
  --- /home/runner/work/webp-imageio/webp-imageio/webp-imageio/api/webp-imageio.api
  +++ /home/runner/work/webp-imageio/webp-imageio/webp-imageio/build/api/webp-imageio.api
  @@ -1,6 +1,4 @@
   public final class com/luciad/imageio/webp/CompressionType : java/lang/Enum {
  -	public static final field Lossless Lcom/luciad/imageio/webp/CompressionType;
  -	public static final field Lossy Lcom/luciad/imageio/webp/CompressionType;
   	public static fun getEntries ()Lkotlin/enums/EnumEntries;
   	public static fun valueOf (Ljava/lang/String;)Lcom/luciad/imageio/webp/CompressionType;
   	public static fun values ()[Lcom/luciad/imageio/webp/CompressionType;
```

Running `apiDump` generates dump without enum entries.

Expected result: Enum entries are part of api dump

This happened in: https://github.com/usefulness/webp-imageio/pull/212 - a project with no custom config. I can try to come up with simpler repro if needed 👀 
I checked release notes, but this is not a multiplatform project, it doesn't have any custom config, the file name name did not change and is [all-lowercase](https://github.com/usefulness/webp-imageio/blob/master/webp-imageio/api/webp-imageio.api)

It seems like the dump is generated properly if I remove or make the companion object `public`

fzhinkin:
#245 fixed the way `const var`s from companion objects are handled. Unfortunately, the solution interferes with how any other static final fields are processed.

This PR postpones companion class resolution until it's known that a field corresponds to a property declared in a companion object.

Fixes #250

### DIFF
--- a/src/main/kotlin/api/KotlinSignaturesLoading.kt
+++ b/src/main/kotlin/api/KotlinSignaturesLoading.kt
@@ -144,7 +144,7 @@ private fun FieldNode.buildFieldSignature(
             /*
              * In certain cases (like with enum entries), the fact a field is both static and final does not
              * imply it belongs to a companion object.
-             * We don't update field's the `companionClass` until we're 100% sure the field was
+             * We don't update field's `companionClass` until we're 100% sure the field was
              * actually moved from the companion.
              */
             companionClass = companionClassCandidate
--- a/src/main/kotlin/api/KotlinSignaturesLoading.kt
+++ b/src/main/kotlin/api/KotlinSignaturesLoading.kt
@@ -127,11 +127,11 @@ private fun FieldNode.buildFieldSignature(
         foundAnnotations.addAll(companionClass?.visibleAnnotations.orEmpty())
         foundAnnotations.addAll(companionClass?.invisibleAnnotations.orEmpty())
     } else if (isStatic(access) && isFinal(access)) {
-        companionClass = ownerClass.companionName(ownerClass.kotlinMetadata)?.let {
+        val companionClassCandidate = ownerClass.companionName(ownerClass.kotlinMetadata)?.let {
             classes[it]
         }
 
-        val property = companionClass?.kmProperty(name)
+        val property = companionClassCandidate?.kmProperty(name)
 
         if (property != null && JvmFlag.Property.IS_MOVED_FROM_INTERFACE_COMPANION(property.flags)) {
             /*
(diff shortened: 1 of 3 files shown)


src/main/kotlin/api/KotlinSignaturesLoading.kt
```diff
@@ -140,7 +140,14 @@ private fun FieldNode.buildFieldSignature(
...
              */
-            foundAnnotations.addAll(companionClass!!.methods.annotationsFor(property.syntheticMethodForAnnotations))
+            foundAnnotations.addAll(companionClassCandidate.methods.annotationsFor(property.syntheticMethodForAnnotations))
+            /*
+             * In certain cases (like with enum entries), the fact a field is both static and final does not
+             * imply it belongs to a companion object.
+             * We don't update field's `companionClass` until we're 100% sure the field was
+             * actually moved from the companion.
+             */
+            companionClass = companionClassCandidate
```
martinbonnin, line 150: What about not updating `companionClass` at all here? Looks like `CompanionFieldBinarySignature` is only to be used for the actual field referencing the companion? 
The bug is me copy/pasting the code from the other branch without realising there would be a side effect at the end of the function :/

fzhinkin: In general, it seems reasonable to include `companionClass` so its visibility can be taken into account. But this part seems to be totally broken.

I'll remove the `companionClass` update and open a separate issue on proper `const val` handling.

> The bug is me copy/pasting the code from the other branch without realising there would be a side effect at the end of the function :/

No worries, it's all about the way JVM ABI is built and filtered (to be more precise, how complicated it became at some point) in conjunction with the test coverage not being full.

PR #255, opened 2024-07-12
Update kotlin-metadata-jvm

fzhinkin:
Replaced `kotlinx-metadata-jvm` with `kotlin-metadata-jvm,` and updated the code to use a new API.

### DIFF
--- a/src/main/kotlin/api/KotlinMetadataVisibilities.kt
+++ b/src/main/kotlin/api/KotlinMetadataVisibilities.kt
@@ -67,7 +67,7 @@ internal val ClassNode.kotlinMetadata: KotlinClassMetadata?
                 extraInt = get("xi") as Int?
             )
         }
-        return KotlinClassMetadata.readStrict(header)
+        return KotlinClassMetadata.readLenient(header)
     }
 
 
--- a/src/main/kotlin/api/KotlinMetadataVisibilities.kt
+++ b/src/main/kotlin/api/KotlinMetadataVisibilities.kt
@@ -5,20 +5,18 @@
 
 package kotlinx.validation.api
 
-import kotlinx.metadata.*
-import kotlinx.metadata.jvm.*
-import kotlinx.metadata.jvm.KotlinClassMetadata.Companion.COMPATIBLE_METADATA_VERSION
+import kotlin.metadata.*
+import kotlin.metadata.jvm.*
 import org.objectweb.asm.tree.*
 
 internal class ClassVisibility(
     val name: String,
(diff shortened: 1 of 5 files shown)


src/main/kotlin/api/KotlinSignaturesLoading.kt
```diff
@@ -132,10 +131,9 @@ private fun FieldNode.buildFieldSignature(
         }
 
         val property = companionClassCandidate?.kmProperty(name)
-
-        if (property != null && JvmFlag.Property.IS_MOVED_FROM_INTERFACE_COMPANION(property.flags)) {
```
fzhinkin, line 136: That was a bug, and `JvmFlag.Property.IS_MOVED_FROM_INTERFACE_COMPANION` worked here purely by coincidence.

src/main/kotlin/api/KotlinMetadataVisibilities.kt
```diff
@@ -68,15 +67,7 @@ internal val ClassNode.kotlinMetadata: KotlinClassMetadata?
...
-        return KotlinClassMetadata.read(header)
-            ?: error(
-                """
-                Incompatible version of Kotlin metadata.
-                Maximal supported Kotlin metadata version: ${COMPATIBLE_METADATA_VERSION.joinToString(".")},
-                $name Kotlin metadata version: ${header.metadataVersion.joinToString(".")}.
-                As a workaround, it is possible to manually update 'kotlinx-metadata-jvm' version in your project.
-            """.trimIndent()
-            )
+        return KotlinClassMetadata.readStrict(header)
```
sandwwraith, line 70: I think it is possible here to use `readLenient()` — we added this function specifically with BCV in mind. See the difference explained here: https://github.com/JetBrains/kotlin/blob/master/libraries/kotlinx-metadata/jvm/ReadMe.md#working-with-different-versions

fzhinkin: Yeah, it makes perfect sense

src/main/kotlin/api/KotlinSignaturesLoading.kt
fzhinkin: Thanks to the new stable API, it became obvious! ;)

shanshin (approved):

fzhinkin:
kotlin-metadata-jvm seems to be API-compatible with Kotlin 1.6, but the API version is not set explicitly. Here's an issue on that: https://youtrack.jetbrains.com/issue/KT-69903

PR #256, opened 2024-07-15
Prototype for project isolation

shanshin:


src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -371,10 +384,21 @@ private class KlibValidationPipelineBuilder(
         val klibMergeInferredDir = projectBuildDir.flatMap { pd -> klibInferDumpConfig.apiDir.map { pd.resolve(it) } }
         val klibExtractedFileDir = klibMergeInferredDir.map { it.resolve("extracted") }
 
+        val runtimeClasspath = project.prepareKlibValidationClasspath()
+
         val klibMerge = project.mergeKlibsUmbrellaTask(klibDumpConfig, klibMergeDir)
+        klibMerge.configure {
+            it.runtimeClasspath.from(runtimeClasspath)
+        }
```
ALikhachev, line 392: Nit: do this inside the task registering function (here and below)?

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -629,3 +658,56 @@ private val Project.jvmDumpFileName: String
...
+    val dependencyConfiguration =
+        project.configurations.create("bcv-rt-klib-cp") {
+            it.description = "Runtime classpath for running binary-compatibility-validator."
+            it.isCanBeResolved = false
+            it.isCanBeConsumed = false
+            it.isCanBeDeclared = true
+            it.isVisible = false
+
+            it.defaultDependencies { dependencySet ->
+                dependencySet.add(project.dependencies.create("org.jetbrains.kotlin:kotlin-compiler-embeddable:${KotlinVersion.CURRENT}"))
```
ALikhachev, line 672: Here too, and here it is even more critical as it configures an incorrect compiler version. For example, on Gradle 7.6, it will configure 1.7.10

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -629,3 +658,56 @@ private val Project.jvmDumpFileName: String
     get() = "$name.api"
 private val Project.klibDumpFileName: String
     get() = "$name.klib.api"
+
+private fun Project.prepareKlibValidationClasspath(): NamedDomainObjectProvider<Configuration> {
+    val dependencyConfiguration =
+        project.configurations.create("bcv-rt-klib-cp") {
```
ALikhachev, line 664: have you considered using `detachedConfiguration`? I see no big problems in using regulars ones, but detached would allow. to hide these configurations from users completely

src/main/kotlin/KotlinApiBuildTask.kt
```diff
@@ -42,13 +46,37 @@ public abstract class KotlinApiBuildTask @Inject constructor(
     @get:PathSensitive(PathSensitivity.RELATIVE)
     public abstract val inputDependencies: ConfigurableFileCollection
 
+    @get:Inject
+    public abstract val executor: WorkerExecutor
```
ALikhachev, line 50: I would make move it up to the parent class

ALikhachev:
In general LGTM, thanks!

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -629,3 +658,56 @@ private val Project.jvmDumpFileName: String
...
+            it.description = "Runtime classpath for running binary-compatibility-validator."
+            it.isCanBeResolved = false
+            it.isCanBeConsumed = false
+            it.isCanBeDeclared = true
+            it.isVisible = false
+            it.defaultDependencies { dependencySet ->
+                dependencySet.add(project.dependencies.create("org.ow2.asm:asm:9.6"))
+                dependencySet.add(project.dependencies.create("org.ow2.asm:asm-tree:9.6"))
+
+                if (KotlinVersion.CURRENT.major >= 2) {
```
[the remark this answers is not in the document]
shanshin: Unfortunately, we can't use `getKotlinPluginVersion` because BCV plugin and KGP may be placed in different classloaders.
I implemented workaround.

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
shanshin: dependency providers can't be used in detachedConfiguration, so I left regular one

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -683,25 +687,27 @@ private fun Project.prepareKlibValidationClasspath(): NamedDomainObjectProvider<
...
-                } else {
-                    // for older 0.6.2 kotlin versions (< 2.0.0) use nearest metadata library
-                    dependencySet.add(project.dependencies.create("org.jetbrains.kotlin:kotlin-metadata-jvm:2.0.0"))
-                }
-            }
         }
+
+    project.dependencies.add(dependencyConfiguration.name, "org.ow2.asm:asm:9.6")
+    project.dependencies.add(dependencyConfiguration.name, "org.ow2.asm:asm-tree:9.6")
+    project.dependencies.addProvider(dependencyConfiguration.name, metadataDependencyVersion.map { version -> "org.jetbrains.kotlin:kotlin-metadata-jvm:$version" })
```
ALikhachev, line 709: This method is available since Gradle 7.0, but the README states
> Binary compatibility validator plugin requires Gradle 6.0 or newer.

But I guess that's fine for now. We in KGP going to drop support for Gradle 6 soon anyway.

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -711,3 +717,60 @@ private fun Project.prepareJvmValidationClasspath(): NamedDomainObjectProvider<C
...
+    plugins.withId("org.jetbrains.kotlin.multiplatform") {
+        block(readVersion())
+    }
+    plugins.withId("org.jetbrains.kotlin.android") {
+        block(readVersion())
+    }
+}
+
+
+private fun Project.readVersion(): String {
```
ALikhachev, line 744: Maybe it makes sense to introduce some kind of caching to not read it each time. We in KGP just store it as static string: https://github.com/JetBrains/kotlin/blob/a2d9a1926df32cd8f107bb41adefe3cbd33db48d/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/plugin/KotlinPluginWrapper.kt#L406

ALikhachev (approved):

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -629,3 +658,119 @@ private val Project.jvmDumpFileName: String
...
+    val extension = extensions.findByName("kotlin")
+    if (extension == null) {
+        logger.warn("Binary compatibility plugin could not find Kotlin extension")
+        return "2.0.0"
```
ALikhachev, line 748: Nit: you can return null, then the convention value will be used (which is already `2.0.0`)

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
shanshin: I doubt that using multiple Kotlin plugins in one project is a common case.

ALikhachev: I mean, it may affect configuration time performance in the case of applying BCV to every module of a large build

ALikhachev: However, it may be not so easy to cache it for this case correctly. You may face different versions of KGP in different projects

shanshin: Yep, because I need cache map, like Project -> Kotlin Version.
And it will require the same number of accesses to resources.

Issue #261, opened 2024-07-24
BCV checks KLib ABIs under `build` dirs


Goooler:
Enable KLib ABI validation with the version `0.16.2` and dump KLib ABIs, `apiCheck` works on my local device but CI:

```
* What went wrong:
Execution failed for task ':compose-multiplatform:library:klibApiCheck'.
> API check failed for project library.

  --- /home/runner/work/Calendar/Calendar/compose-multiplatform/library/build/api/klib-all/extracted/library.klib.api
  +++ /home/runner/work/Calendar/Calendar/compose-multiplatform/library/build/api/klib/library.klib.api
  @@ -5,7 +5,7 @@
   // - Show manifest properties: true
   // - Show declarations: true
   
  -// Library unique name: <kizitonwose-calendar.compose-multiplatform:library>
  +// Library unique name: <Calendar.compose-multiplatform:library>
   final enum class com.kizitonwose.calendar.compose.heatmapcalendar/HeatMapWeekHeaderPosition : kotlin/Enum<com.kizitonwose.calendar.compose.heatmapcalendar/HeatMapWeekHeaderPosition> { // com.kizitonwose.calendar.compose.heatmapcalendar/HeatMapWeekHeaderPosition|null[0]
       enum entry End // com.kizitonwose.calendar.compose.heatmapcalendar/HeatMapWeekHeaderPosition.End|null[0]
       enum entry Start // com.kizitonwose.calendar.compose.heatmapcalendar/HeatMapWeekHeaderPosition.Start|null[0]
  @@ -364,6 +364,12 @@
       final fun toString(): kotlin/String // com.kizitonwose.calendar.data/WeekDateRange.toString|toString(){}[0]
   }
```

You can repro this issue at https://github.com/kizitonwose/Calendar/pull/566.

fzhinkin:
@Goooler, the current klib-abi-validation pipeline extracts an ABI for supported targets from a file stored within the source's api-dir and writes a resulting file into `build/api/klib-all/extracted/...`. 

In your example, the first file is a file created by filtering out unsupported targets from an ABI file stored within project sources, and the second file is an ABI generated for the current revision:
```
  --- /home/runner/work/Calendar/Calendar/compose-multiplatform/library/build/api/klib-all/extracted/library.klib.api
  +++ /home/runner/work/Calendar/Calendar/compose-multiplatform/library/build/api/klib/library.klib.api
```

So there's nothing wrong in both files being stored in build dir.


[...]

PR #262, opened 2024-07-24
Avoid using project.name for ignored projects check

### LINKED ISSUE #16: Allow ignoring project by path
I have a project where multiple modules have the same name but different path (`:feature1:sample`, `:feature2:sample`) and currently I cannot ignore just one of them since you specify the name of the module which in both cases is `sample`. 

This could be done in a backwards compatible way by checking if the ignored project starts with `:` and if so `project.path` is used instead.

### LINKED ISSUE #257: Subprojects that contain the same name will be ignored
Declare subprojects in `settings.gradle` like:

```groovy
include(
    ":sample",
    ":compose-multiplatform:library",
    ":compose-multiplatform:sample"
)
``` 

the structure should be:

```sh
├── compose-multiplatform
│   ├── library
│   └── sample
├── sample
```

then configure BCV in root `build.gradle`:

```kotlin
plugins {
    id("org.jetbrains.kotlinx.binary-compatibility-validator:0.15.1")
}

apiValidation {
    ignoredProjects += listOf(
        "sample"
    )
}
```

then dump APIs for them, `compose-multiplatform:sample` and `sample` projects are both ignored.

If we add `compose-multiplatform:sample` into ignoredProjects and run again, there will be an error:

```
Cannot find excluded project compose-multiplatform:sample in all projects: [root-project, compose-multiplatform, core, sample, library, sample]
```

as you can see, there are 2 projects listed here with the same name `sample`, they should be split.

Goooler:
- Closes #16.
- Closes #257.

> The project's name is not necessarily unique within a project hierarchy. You should use the getPath() method for a unique identifier for the project.

See https://docs.gradle.org/current/javadoc/org/gradle/api/Project.html#getName().

### DIFF
--- a/src/functionalTest/kotlin/kotlinx/validation/test/SubprojectsWithPluginOnRootTests.kt
+++ b/src/functionalTest/kotlin/kotlinx/validation/test/SubprojectsWithPluginOnRootTests.kt
@@ -14,6 +14,7 @@ import kotlinx.validation.api.runner
 import kotlinx.validation.api.test
 import org.assertj.core.api.Assertions
 import org.junit.Test
+import kotlin.test.assertContains
 import kotlin.test.assertTrue
 
 internal class SubprojectsWithPluginOnRootTests : BaseKotlinGradleTest() {
@@ -315,6 +316,41 @@ internal class SubprojectsWithPluginOnRootTests : BaseKotlinGradleTest() {
             val apiSub2 = rootProjectDir.resolve("sub2/$API_DIR/sub2.api")
             assertTrue(apiSub2.exists(), "api dump file ${apiSub2.path} should exist")
             Assertions.assertThat(apiSub2.readText()).isEqualToIgnoringNewLines("")
+        }
+    }
+
+    /**
+     * https://github.com/Kotlin/binary-compatibility-validator/issues/257
+     */
+    @Test
+    fun `using project name instead of path should not be ignored`() {
+        val runner = test {
+            createProjectHierarchyWithPluginOnRoot()
+            rootProjectDir.resolve("build.gradle.kts").writeText(
+                """
+                apiValidation {
+                    ignoredProjects += listOf(
+                        "subsub1"
+                    )
+                }
+                """.trimIndent()
+            )
+
+            runner {
+                arguments.add(":apiCheck")
+            }
+        }
+
+        try {
+            runner.build()
+            error("Should have failed.")
+        } catch (t: Throwable) {
+            assertContains(
+                t.stackTraceToString(),
+                """
+                    Cannot find excluded project subsub1 in all projects: [:, :sub1, :sub2, :sub1:subsub1, :sub1:subsub2]
+                """.trimIndent()
+            )
         }
     }
 }
--- a/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
+++ b/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
@@ -17,7 +17,6 @@ import org.jetbrains.kotlin.gradle.plugin.*
 import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
 import org.jetbrains.kotlin.konan.target.HostManager
 import org.jetbrains.kotlin.library.abi.ExperimentalLibraryAbiReader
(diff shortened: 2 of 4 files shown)


Goooler:
2 tests failed, seems they are related to #261.

`kotlinx.validation.test.MultiPlatformSingleJvmTargetTest.testApiCheckFails`:

```
* What went wrong:
Execution failed for task ':jvmApiCheck'.
> API check failed for project root project 'testproject'.
  --- /private/var/folders/ry/cppncsdj3rd634f18p7m2n000000gn/T/junit8646780838980940615/api/testproject.api
  +++ /private/var/folders/ry/cppncsdj3rd634f18p7m2n000000gn/T/junit8646780838980940615/build/api/testproject.api
  @@ -1,9 +1,9 @@
  -public final class com/company/subsub2/Subsub2Class {
  +public final class com/company/subsub1/Subsub1Class {
   	public fun <init> ()V
   	public final fun getProp ()I
   }
   
  -public final class com/company/subsub1/Subsub1Class {
  +public final class com/company/subsub2/Subsub2Class {
   	public fun <init> ()V
   	public final fun getProp ()I
   }
  
  You can run :root project 'testproject':apiDump task to overwrite API declarations
```


`kotlinx.validation.test.MultipleJvmTargetsTest.testApiCheckFails`

```
* What went wrong:
Execution failed for task ':jvmApiCheck'.
> API check failed for project root project 'testproject'.
  --- /private/var/folders/ry/cppncsdj3rd634f18p7m2n000000gn/T/junit5952170624883697109/api/jvm/testproject.api
  +++ /private/var/folders/ry/cppncsdj3rd634f18p7m2n000000gn/T/junit5952170624883697109/build/api/jvm/testproject.api
  @@ -1,9 +1,9 @@
  -public final class com/company/subsub2/Subsub2Class {
  +public final class com/company/subsub1/Subsub1Class {
   	public fun <init> ()V
   	public final fun getProp ()I
   }
   
  -public final class com/company/subsub1/Subsub1Class {
  +public final class com/company/subsub2/Subsub2Class {
   	public fun <init> ()V
   	public final fun getProp ()I
   }
  
  You can run :root project 'testproject':apiDump task to overwrite API declarations
```

fzhinkin:
@Goooler thanks for taking care of the problem!

While paths are better for identifying projects, there's already plethora of projects using BCV and relying on `ignoredProjects` property.
We can follow the suggestion from #16 on how to preserve compatibility, and treat each of `ignoredProjects` values either as a project name, or as a project path depending on the first character (assuming that only the project path could start with `:`). 
Additionally, it would nice to print a warning for every `ignoredProjects` entry that is no a project path (so that users gradually migrate to project paths and we can drop project names support in the future).

Goooler:
This might be a breaking change, but it would be worth breaking in the next release, users who want to update should update their `ignoredProjects` instead of treating it as a warning.

Yeah, we can remove `:` prefixes for path checks, but there will be more complex cases for us to treat, such as the root project, path as `:`. I suggest we should retain the original paths for checks.

fzhinkin:
> This might be a breaking change, but it would be worth breaking in the next release, users who want to update should update their `ignoredProjects` instead of treating it as a warning.

It's better to provide a smooth and graceful migration path when possible. Those who already use `ignoredProjects` are fine with how it works. They'll have to switch to paths at some point, but it should not block them from updating a BCV. 
In some cases, a breaking change is required to fix a problem that could not be fixed otherwise. But in case with path/name checks, behavior could be updated smoothly, without breaking what already work. 

We just need to transform
```
if (project.path in extension.ignoredProjects) { ... }
```
into something like
```
fun Project.isIgnored(extension.ignoredProjects: ApiValidationExtension) = path in extension.ignoredProjects || name in extension.ignoredProjects

if (project.isIgnored(extension)) { ... }
```
And also update `Project.validateExtension` to check that `ignoredProjects` values are either existing names, or paths (printing a warning for names along the way).

shanshin:
As part of [the migration of separate BCV functionality to Kotlin Gradle Plugin](https://youtrack.jetbrains.com/issue/KT-71098/Migrate-Binary-Compatibility-Validator-to-Kotlin-Gradle-Plugin), the addition of new features to a separate plugin has been discontinued.

PR #264, opened 2024-07-27
Add min Gradle version checker

### LINKED ISSUE #263: Error when using version 0.16.2 in projects with a Gradle version older than 8.2
The error is:

```
* What went wrong:
'void org.gradle.api.artifacts.Configuration.setCanBeDeclared(boolean)'
```

I suppose the problem was introduced in the last release, specifically because of using `configuration.isCanBeDeclared` in this file: https://github.com/Kotlin/binary-compatibility-validator/blob/14ebd0231c33a8e6c66414c95ca1226b09897b57/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt#L662

`setCanBeDeclared` was introduced in Gradle 8.2

Goooler:
Closes #263.

### DIFF
--- a/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
+++ b/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
@@ -11,6 +11,7 @@ import org.gradle.api.artifacts.Configuration
 import org.gradle.api.plugins.*
 import org.gradle.api.provider.*
 import org.gradle.api.tasks.*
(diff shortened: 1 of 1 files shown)


fzhinkin (changes requested):
To fix #263 a problematic method should be invoked only if project‘s Gradle supports it. Bumping up a minimum required Gradle version is not an option as a lot of projects use 7.x.

Goooler:
Addressing another fix to #265.

PR #265, opened 2024-07-27
Compat isCanBeDeclared usages

Goooler:
Close #263.

PR #266, opened 2024-07-31
Fixed support for Gradle < 8.2

### LINKED ISSUE #263: Error when using version 0.16.2 in projects with a Gradle version older than 8.2

shanshin:
Fixes #263

### DIFF
--- a/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
+++ b/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
@@ -16,7 +16,6 @@ import org.jetbrains.kotlin.gradle.plugin.*
 import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
 import org.jetbrains.kotlin.konan.target.HostManager
 import org.jetbrains.kotlin.library.abi.ExperimentalLibraryAbiReader
-import org.jetbrains.kotlin.library.abi.LibraryAbiReader
 import java.io.*
 import java.util.*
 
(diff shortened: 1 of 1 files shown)


src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -648,18 +647,32 @@ private val Project.jvmDumpFileName: String
...
 
+private fun Project.isGradle8Dot2OrHigher(): Boolean {
+    val parts = gradle.gradleVersion.split('.')
+    if (parts.size < 2) return false
+    val major = parts[0].toIntOrNull() ?: return false
+    val minor = parts[1].toIntOrNull() ?: return false
+    if (major < 8) return false
+    if (major == 8 && minor < 2) return false
+    return true
+}
```
JakeWharton, line 658: See #265 which uses Gradle's [`GradleVersion` API](https://docs.gradle.org/current/javadoc/org/gradle/util/GradleVersion.html) that obviates the need to do any manual version parsing like this.

shanshin: Indeed, thanks, missed that PR

PR #271, opened 2024-10-10
Filter out `$suspendImpl` functions from dumps

### LINKED ISSUE #270: Filter out `$suspendImpl` functions declared in interfaces
On JVM, the compiler generates special `<original function name>$suspendImpl` methods for every open suspendable function. For classes, these methods are package private and are never included in dump. For interfaces, however, such functions could only be either private (but we need to call them outside the interface, so that doesn't work), or public (we are here now). So, these public functions do not constitute the public ABI and are public only due to restrictions imposed on class files.

BCV should filter these methods out of generated dumps.

fzhinkin:
Closes #270

### DIFF
--- a/src/main/kotlin/api/KotlinMetadataSignature.kt
+++ b/src/main/kotlin/api/KotlinMetadataSignature.kt
@@ -60,6 +60,7 @@ internal data class MethodBinarySignature(
         super.isEffectivelyPublic(classAccess, classVisibility)
                 && !isAccessOrAnnotationsMethod()
                 && !isDummyDefaultConstructor()
+                && !isSuspendImplMethod()
 
     override fun findMemberVisibility(classVisibility: ClassVisibility?): MemberVisibility? {
         return super.findMemberVisibility(classVisibility)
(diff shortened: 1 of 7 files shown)


src/main/kotlin/api/KotlinMetadataSignature.kt
```diff
@@ -71,6 +72,14 @@ internal data class MethodBinarySignature(
...
     private fun isDummyDefaultConstructor() =
         access.isSynthetic && name == "<init>" && desc == "(Lkotlin/jvm/internal/DefaultConstructorMarker;)V"
+
+    /**
+     * Kotlin compiler emits special `<originalFunctionName>$suspendImpl` methods for open
+     * suspendable functions. These synthetic functions could only be invoked from original function,
+     * or from a corresponding continuation. They don't constitute class's public ABI, but in some cases
+     * they might be declared public (namely, in case of default interface methods).
+     */
+    private fun isSuspendImplMethod() = access.isSynthetic && name.endsWith("\$suspendImpl")
```
lukellmann, line 82: It seems like the `$suspendImpl` suffix is missing when `@JvmName` is involved, like with [`ChannelIterator.next0`](https://github.com/Kotlin/kotlinx.coroutines/blob/6c6df2b850382887462eeaf51f21f58bd982491d/kotlinx-coroutines-core/common/src/channels/Channel.kt#L596): https://github.com/Kotlin/kotlinx.coroutines/pull/4244/files#diff-339272aa30544e4ab13c2daaf50891927916f46915a92419bb04d8d6ef6cdd23L701-R706

fzhinkin: @ilmirus is there a reason why `$suspendImpl` methods with a fancy jvm-name don't get their suffix?

fzhinkin: Yeap, `@JvmName` on open methods is not something really supported (thus `@Suppress("INAPPLICABLE_JVM_NAME")`), so I don't think it is not worth bothering on the BCV end either.

After all, users who consciously opted it in by suppressing `INAPPLICABLE_JVM_NAME` can also explicitly filter out it from a dump, if needed.

shanshin (approved):

PR #284, opened 2024-12-18
Update ASM to 9.7 to support class files generated for Java 23

fzhinkin:


fzhinkin:
@shanshin and as you envisioned, compatibility tests are pretty useless when it comes to catching these errors.

shanshin (approved):

shanshin:
@fzhinkin, do you know if there is a new version of the `org.jetbrains.intellij.deps:asm-all` artifact?

fzhinkin:
Nope, `9.6.1` is the latest. But class file version check is disabled there, so it works with class files from JDK23+ despite being a bit outdated.

JakeWharton:
This also should support JDK 24, too!

dalewking:
Just happened to notice that while version 0.17.0 supposedly contains this change to switch to newer asm, that does not appear to be the case. I am using 0.17.0 of BCV and it says that we are using 9.6 of asm. Here is a report from dependencyUpdates task:

```
> Task :dependencyUpdates

------------------------------------------------------------
: Project Dependency Updates (report to plain text file)
------------------------------------------------------------

The following dependencies are using the latest milestone version:
...
 - org.jetbrains.kotlinx.binary-compatibility-validator:org.jetbrains.kotlinx.binary-compatibility-validator.gradle.plugin:0.17.0
...

The following dependencies have later milestone versions:
...
 - org.ow2.asm:asm [9.6 -> 9.7.1]
     http://asm.ow2.io/
 - org.ow2.asm:asm-tree [9.6 -> 9.7.1]
     http://asm.ow2.io/
```

Doesn't really make a difference to me since we are using JDK 17, but was checking what dependencies we could upgrade and was trying to figure out where the asm dependency was coming from

PR #307, opened 2025-07-08
Update cross-compilation check for 2.2.20

fzhinkin:
Starting from 2.2.20, cross-compilation is enabled by default (see https://youtrack.jetbrains.com/issue/KT-76421). This PR modifies corresponding check to reflect that.

### DIFF
--- a/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
+++ b/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
@@ -494,6 +494,8 @@ private class KlibValidationPipelineBuilder(
             it.dumps.addAll(inferredDumps)
         }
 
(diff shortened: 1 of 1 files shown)


shanshin (approved):

src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -494,6 +494,8 @@ private class KlibValidationPipelineBuilder(
             it.dumps.addAll(inferredDumps)
         }
 
+        val kgpVersion = readVersion()
```
fzhinkin, line 497: @shanshin while looking through the code, I figured out that that should be a better way to read a KGP version, so I switched to it.

